PR #93, opened 2022-07-27
Customized Keycloak Image

overheadhunter:
This adds a new folder `/keycloak` which is used to build a [custom keycloak image](https://www.keycloak.org/server/containers).

The custom image is pre-configured for being served under `/kc` and to connect to a postgres database.

Furthermore it contains a themed login page that looks like this and needs to be set in the `realm.json`(WIP):

Screenshot 2022-07-27 at 16 10 41


overheadhunter:
Screenshot 2022-08-16 at 17 44 32

overheadhunter:
We might need a different source for social login icons (`<i class="fa fa-github"></i>`), since we don't want to include `node_modules/patternfly/dist/css/patternfly.min.css` (from the parent "keycloak" theme) due to its size.

@tobihagemann any suggestion for some other icon source? maybe a tailwind extension? we can specify class names ourselves.

tobihagemann:
I would suggest to include Font Awesome directly if possible. I'm not aware of a different source that provides brand icons that extensive and easy.

PR #105, opened 2022-08-05
Add tests for the remote user puller

SailReal:


backend/src/main/java/org/cryptomator/hub/RemoteUserPuller.java
```diff
@@ -30,17 +30,38 @@ void sync() {
...
+	}
+
+	//visible for testing
+	void sync(Map<String, Group> keycloakGroups, Map<String, User> keycloakUsers, Map<String, Group> databaseGroups, Map<String, User> databaseUsers) {
+		syncAddedUsers(keycloakUsers, databaseUsers);
+		var deletedUsers = syncDeletedUsers(keycloakUsers, databaseUsers);
+		syncUpdatedUsers(keycloakUsers, databaseUsers, deletedUsers);
+		syncAddedGroups(keycloakGroups, databaseGroups);
+		var deletedGroups = syncDeletedGroups(keycloakGroups, databaseGroups);
+		syncUpdatedGroups(keycloakGroups, databaseGroups, deletedGroups);
```
overheadhunter, line 43: would generics help to avoid duplication of these three methods?

backend/src/test/java/org/cryptomator/hub/RemoteUserPullerTest.java
```diff
@@ -0,0 +1,277 @@
...
+				"foo,bar,baz;,;foo,bar,baz",
+				"foo,bar,baz;la,li,lu;foo,bar,baz",
+				"foo,bar,baz;la,bar,baz;foo",
+				"baz;foo,bar,baz;,",
+				",;foo,bar,baz;,",
+				",;,;,"
+		}, delimiterString = ";")
+		public void testAddUsers(@ConvertWith(StringArrayConverter.class) String[] keycloakUserIdString, @ConvertWith(StringArrayConverter.class) String[] databaseUserIdString, @ConvertWith(StringArrayConverter.class) String[] addedUserIdString) {
+			Map<String, User> keycloakUsers = Mockito.mock(Map.class);
+			Map<String, User> databaseUsers = Mockito.mock(Map.class);
```
overheadhunter, line 54: wouldn't it be easier to create Spy objects around "real" maps?

backend/src/test/java/org/cryptomator/hub/RemoteUserPullerTest.java
```diff
@@ -0,0 +1,277 @@
...
+			}
+		}
+	}
+
+	private static class StringArrayConverter extends SimpleArgumentConverter {
+
+		@Override
+		protected Object convert(Object source, Class<?> targetType) throws ArgumentConversionException {
+			if (source instanceof String && String[].class.isAssignableFrom(targetType)) {
+				return ((String) source).split(",");
```
overheadhunter, line 269:
```suggestion
			if (source instanceof String s && String[].class.isAssignableFrom(targetType)) {
				return s.split(",");
```

backend/src/main/java/org/cryptomator/hub/RemoteUserPuller.java
SailReal: Yes, a lot :heart_eyes:, implemented in https://github.com/cryptomator/hub/pull/105/commits/e6c560df17cd38797ecb6c2d3ba7330b950cfb13

backend/src/test/java/org/cryptomator/hub/RemoteUserPullerTest.java
SailReal: Hmm I'm not sure, it's right that we don't want to test the map type itself but I find the initialization of the keys and values in the map easier and more comprehensible but I can change it if you want :)

overheadhunter (approved):

PR #114, opened 2022-08-10
Fixed history mode

overheadhunter:
Making the backend work with "History Mode" routing was a PITA and very fragile. When building native images with GraalVM, due to slight behavioural differences, the servlet filter responsible for routing every 404 response to the index.html failed to initialize with obscure error messages (`Failed to load config value of type class java.lang.String for: quarkus.resteasy.path`).

Furthermore, waiting for a 404 response means that the response was already written, leading to a very sad attempt to try and rewrite it: Status code and (Content Type) headers were already sent, leading browsers to treat the index.html as plaintext. In some browsers (didn't verify if this is consistent behaviour) non-html content leads to CSP errors.

This PR inverts the matching logic: Instead of "all but /api/* needs to be routed by the frontend router", we created a dedicated subresource called `/app/*` for frontend routes.

While the index.html is still generated at `/`, the Vue router is configured with `/app` as its base path.

This simplifies the VueHistoryModeFilter, as it can now rely on "positive" matching rules without exceptions. The http status code and content type is now correctly determined as if the user agent asked for index.html directly.

### DIFF
--- a/frontend/src/common/config.ts
+++ b/frontend/src/common/config.ts
@@ -1,6 +1,6 @@
 import AxiosStatic from 'axios';
 
-export const frontendBaseURL = `${location.protocol}//${location.host}${import.meta.env.BASE_URL}`;
+export const frontendBaseURL = `${location.protocol}//${location.host}/app`;
 export const backendBaseURL = import.meta.env.DEV
   ? 'http://localhost:8080/api/'
   : new URL('/api/', location.href).href;
--- a/backend/src/main/java/org/cryptomator/hub/filters/VueHistoryModeFilter.java
+++ b/backend/src/main/java/org/cryptomator/hub/filters/VueHistoryModeFilter.java
@@ -1,7 +1,5 @@
 package org.cryptomator.hub.filters;
 
-import org.eclipse.microprofile.config.inject.ConfigProperty;
-
 import javax.servlet.FilterChain;
 import javax.servlet.ServletException;
 import javax.servlet.ServletRequest;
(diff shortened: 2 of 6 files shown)


frontend/src/common/config.ts
```diff
@@ -1,6 +1,6 @@
 import AxiosStatic from 'axios';
 
-export const frontendBaseURL = `${location.protocol}//${location.host}${import.meta.env.BASE_URL}`;
```
infeo, line 3: While i like the change from "one needs to know from context" to something explicit, I cannot say if this is correct. @tobihagemann Can we just replace BASE_URL with a fixed path?

infeo (changes requested):
Nearly everything seems to work.

When I navigate to `https://my.server:XXXX/app` and click the "Go Home" button, i will be redirected to `https://my.server:XXXX/vaults`. There seems a bug in the url resolving when no subresource is given.

frontend/src/common/config.ts
overheadhunter: What this PR doesn't show: I did half a day of experiments with `BASE_URL` and decided to ditch it, as it is not applicable to our app at all. At first I bound the router to the BASE_URL and set the BASE_URL in `vite.config.ts`.

This leads to wrong asset paths, though. We want vite to generate everything at `/` but the router to put everything under `/app`.

overheadhunter: Btw it was me [who introduced the `BASE_URL`](https://github.com/cryptomator/hub/commit/52b1035259495926fbd66492caec2dbb47d4b763#diff-8d302fe8581dca01e60aaf346dfdb0f8e24731551e7e5f91e5da9fc64250f7d2R37), thinking that we might need to customize the apps root. Back then I didn't know what effect this had on the build.

infeo:
> When I navigate to https://my.server:XXXX/app and click the "Go Home" button, i will be redirected to https://my.server:XXXX/vaults. There seems a bug in the url resolving when no subresource is given.

This seems to be a Firefox specific issue, not reproducable with Chromiumbased browsers. I'll open a ticket for it and we can inore it in this PR.

frontend/src/common/config.ts
infeo: Ok, than this change is legit.

infeo (approved):

PR #121, opened 2022-08-12
Show update info in settings screen

### LINKED ISSUE #119: Add Update-Check

### Summary

Add an update-check including a UI indicator for hub.

### Motivation

Prevent usage of outdated software.

### Considered Alternatives

-/-

### Anything else?

DoD:
* Hub instance checks regularly for updates
* If update is available, display it in hub frontend.

overheadhunter:
fixes #119

## TODO
- [x] proper UI
- [x] update Content-Security-Policy header to allow XHRs to api.cryptomator.org in Hub Setup

### DIFF
--- a/frontend/src/common/updatecheck.ts
+++ b/frontend/src/common/updatecheck.ts
@@ -11,7 +11,7 @@ class UpdatesService {
     let config = {
       headers: {
         'Content-Type': 'application/json',
-        'User-Agent': 'Hub/' + localVersion,
+        'X-Hub-Version': localVersion,
         'X-Hub-Instance': 'TODO' //for future uses
       }
     };
--- a/frontend/src/common/updatecheck.ts
+++ b/frontend/src/common/updatecheck.ts
@@ -1,0 +1,21 @@
+import AxiosStatic, { AxiosRequestConfig } from 'axios';
+
+const axiosBaseCfg: AxiosRequestConfig = {
+  headers: {
+    'Content-Type': 'application/json'
+  }
+};
+
+const axios = AxiosStatic.create(axiosBaseCfg);
+
+export class LatestVersionDto {
+  constructor(public stable: string, public beta: string) { }
+}
+
+class UpdatesService {
+  public async get(): Promise<LatestVersionDto> {
+    return axios.get('https://api.cryptomator.org/updates/hub.json').then(response => response.data);
+  }
+}
+
+export const updateChecker = new UpdatesService();
--- a/frontend/src/common/updatecheck.ts
+++ b/frontend/src/common/updatecheck.ts
@@ -1,20 +1,21 @@
-import AxiosStatic, { AxiosRequestConfig } from 'axios';
+import AxiosStatic from 'axios';
 
-const axiosBaseCfg: AxiosRequestConfig = {
-  headers: {
-    'Content-Type': 'application/json'
-  }
-};
-
-const axios = AxiosStatic.create(axiosBaseCfg);
+const axios = AxiosStatic.create();
 
 export class LatestVersionDto {
   constructor(public stable: string, public beta: string) { }
 }
 
 class UpdatesService {
-  public async get(): Promise<LatestVersionDto> {
-    return axios.get('https://api.cryptomator.org/updates/hub.json')
+  public async get(localVersion: string): Promise<LatestVersionDto> {
+    let config = {
+      headers: {
+        'Content-Type': 'application/json',
+        'User-Agent': 'Hub/' + localVersion,
+        'X-Hub-Instance': 'TODO' //for future uses
+      }
+    };
+    return axios.get('https://api.cryptomator.org/updates/hub.json', config)
       .then(response => response.data)
       .catch(err => {
         console.error(err);
--- a/frontend/src/common/updatecheck.ts
+++ b/frontend/src/common/updatecheck.ts
@@ -14,7 +14,18 @@ export class LatestVersionDto {
 
 class UpdatesService {
   public async get(): Promise<LatestVersionDto> {
-    return axios.get('https://api.cryptomator.org/updates/hub.json').then(response => response.data);
+    return axios.get('https://api.cryptomator.org/updates/hub.json')
+      .then(response => response.data)
+      .catch(err => {
+        console.error(err);
+        throw new FetchUpdateError('Unable to get update info.');
+      });
+  }
+}
+
+export class FetchUpdateError extends Error {
+  constructor(msg: string) {
+    super(msg);
   }
 }
 
--- a/frontend/src/components/Settings.vue
+++ b/frontend/src/components/Settings.vue
@@ -107,13 +107,13 @@ const errorOnFetchingUpdates = ref<boolean>(false);
 
 const isBeta = computed(() => semver.prerelease(version.value?.hubVersion ?? '0.1.0') != null);
 const stableUpdateExists = computed(() => {
-  if (version.value && latestVersion.value) {
+  if (version.value && latestVersion.value?.stable) {
     return semver.lt(version.value?.hubVersion , latestVersion.value.stable ?? '0.1.0');
   }
   return false;
 });
 const betaUpdateExists = computed(() => {
-  if (version.value && latestVersion.value) {
+  if (version.value && latestVersion.value?.beta) {
     return semver.lt(version.value?.hubVersion , latestVersion.value.beta ?? '0.1.0-beta1');
   }
   return false;
--- a/frontend/src/components/Settings.vue
+++ b/frontend/src/components/Settings.vue
@@ -108,8 +108,10 @@ onMounted(fetchData);
 async function fetchData() {
   onFetchError.value = null;
   try {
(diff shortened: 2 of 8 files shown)


frontend/src/common/updatecheck.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const axios = AxiosStatic.create(axiosBaseCfg);
+
+export class LatestVersionDto {
+  constructor(public stable: string, public beta: string) { }
+}
+
+class UpdatesService {
+  public async get(): Promise<LatestVersionDto> {
+    return axios.get('https://api.cryptomator.org/updates/hub.json')
```
SailReal, line 17: Shall we put the currently running version in the `User Agent` so we can use that for some statistical analysis?

overheadhunter: And maybe even a hash of the hub installation id?

frontend/src/components/Settings.vue
```diff
@@ -68,25 +85,87 @@
...
+    }
+  }
+}
+
+function checkForUpdates() {
+  stableUpdateExists.value = checkForStableUpdate();
+  if (!stableUpdateExists.value) {
+    betaUpdateExists.value = checkForBetaUpdate();
+  }
+}
```
overheadhunter, line 131: I did some experiments with the `semver` lib. It is very much capable of comparing semver strings (as its name suggests). I.e.

```js
// 1.0.0 < 1.1.0-beta1 < 1.1.0-beta2 < 1.1.0:

it('1.0.0 < 1.1.0-beta1', () => {
  const result = semver.lt('1.0.0', '1.1.0-beta1');
  expect(result).to.be.true;
});

it('1.1.0-beta1 < 1.1.0-beta2', () => {
  const result = semver.lt('1.1.0-beta1', '1.1.0-beta2');
  expect(result).to.be.true;
});

it('1.1.0-beta2 < 1.1.0', () => {
  const result = semver.lt('1.1.0', '1.1.0-beta2');
  expect(result).to.be.false;
});
```

Therefore, there should be no need to manually deconstruct the version string. Regardless of beta or non-beta, just let the semver comparison do its job:

```suggestion
function checkForUpdates() {
  // TODO: can we assert that latestVersion.value != null?
  stableUpdateExists.value = semver.lt(version.value, latestVersion.value.stable);
  betaUpdateExists.value = semver.lt(version.value, latestVersion.value.beta);
}

// TODO: delete other unused functions
```

overheadhunter (commented):
I believe, the logic can be simplified.

frontend/src/components/Settings.vue
```diff
@@ -54,6 +54,23 @@
...
+                  {{ t('settings.update.upToDate.description') }}
+                </p>
+                <p v-else-if="stableUpdateExists" id="version-description" class="inline-flex mt-2 text-sm text-gray-500">
+                  <ExclamationIcon class="shrink-0 text-orange-500 mr-1 h-5 w-5" aria-hidden="true" />
+                  {{ t('settings.update.updateExists.description', [latestVersion.stable]) }}
+                </p>
+                <p v-else id="version-description" class="inline-flex mt-2 text-sm text-gray-500">
+                  <InformationCircleIcon class="shrink-0 text-primary mr-1 h-5 w-5" aria-hidden="true" />
+                  {{ t('settings.update.updateExists.description', [latestVersion.beta]) }}
+                </p>
```
overheadhunter, line 73: Can we differentiate whether we're currently using stable or beta software?

|                          | currently installed: stable | currently installed: beta |
|--------------------------|-----------------------------|---------------------------|
| update available: stable | Update available!           | Update available!         |
| update available: beta   | Up to date (Beta available) | Update available!         |
| update-check failed      | Please check manually!      | Please check manually!    |

frontend/src/components/Settings.vue
infeo: Fixed in https://github.com/cryptomator/hub/pull/121/commits/525339bb199013a669c047e20d6571d574174fb7

frontend/src/common/updatecheck.ts
infeo: Since admin rights are currently needed to access the hubid, but the server info resides inside the settings section, we skip this enhancement for now.

infeo: > currently running version in the User Agent 

Implemented in https://github.com/cryptomator/hub/pull/121/commits/0eaa14ac89c1bd27ee80805c3407844b81ab929b

frontend/src/components/Settings.vue
infeo: Fixed in https://github.com/cryptomator/hub/pull/121/commits/ec84d34858afce90e583e48a61eaa1e91e788d17

frontend/src/components/Settings.vue
```diff
@@ -54,6 +54,27 @@
...
+                  {{ t('settings.update.updateExists.description', [latestVersion.stable]) }}
+                </p>
+                <p v-else-if="betaUpdateExists && isBeta" id="version-description" class="inline-flex mt-2 text-sm text-gray-500">
+                  <ExclamationIcon class="shrink-0 text-orange-500 mr-1 h-5 w-5" aria-hidden="true" />
+                  {{ t('settings.update.updateExists.description', [latestVersion.beta]) }}
+                </p>
+                <p v-else id="version-description" class="inline-flex mt-2 text-sm text-gray-500">
+                  <InformationCircleIcon class="shrink-0 text-primary mr-1 h-5 w-5" aria-hidden="true" />
+                  {{ t('settings.update.updateExists.description', [latestVersion.beta]) }}
+                </p>
```
overheadhunter, line 77: instead of `v-else` I think it's easier to understand this case, if making it an `v-else-if="betaUpdateExists && !isBeta"`.

Or is a `v-else` always required by Vue?

infeo: Fixed in https://github.com/cryptomator/hub/pull/121/commits/73672626c0529ea7fa7b776e2c950f0a36c8e7ec

frontend/src/components/Settings.vue
```diff
@@ -68,25 +89,50 @@
...
 import { Locale } from '../i18n';
+import FetchError from './FetchError.vue';
 
 const { t } = useI18n({ useScope: 'global' });
 
 const version = ref<VersionDto>();
+const isBeta = ref<boolean>(false);
+const latestVersion = ref<LatestVersionDto>();
+const stableUpdateExists = ref<boolean>(false);
+const betaUpdateExists = ref<boolean>(false);
```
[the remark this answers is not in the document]
infeo: This is indeed possible. I changed it to computed properties in https://github.com/cryptomator/hub/pull/121/commits/574e6668c4583584d3819f651a4b81752b5c3c94

tobihagemann (changes requested):

frontend/src/common/updatecheck.ts
```diff
@@ -0,0 +1,33 @@
...
+export class LatestVersionDto {
+  constructor(public stable: string, public beta: string) { }
+}
+
+class UpdatesService {
+  public async get(localVersion: string): Promise<LatestVersionDto> {
+    let config = {
+      headers: {
+        'Content-Type': 'application/json',
+        'User-Agent': 'Hub/' + localVersion,
```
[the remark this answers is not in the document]
tobihagemann: Seems like a very old bug in Chrome: https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name

> **Note**
> The [User-Agent](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent) header is no longer forbidden, [as per spec](https://fetch.spec.whatwg.org/#terminology-headers) — see forbidden header name list (this was implemented in Firefox 43) — it can now be set in a Fetch [Headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers) object, or via XHR [setRequestHeader()](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest#setrequestheader%28%29). However, Chrome will silently drop the header from Fetch requests (see [Chromium bug 571722](https://bugs.chromium.org/p/chromium/issues/detail?id=571722)).

I'm not sure if we should already do this in this PR but I think it's best to move the update checker into the backend.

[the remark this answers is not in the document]
overheadhunter: Ok, let's use `X-Hub-Version`. Strange behaviour, though...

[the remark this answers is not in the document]
overheadhunter: (This will also give us some insights into the actual user agents used)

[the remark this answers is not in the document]
infeo: Changed in https://github.com/cryptomator/hub/pull/121/commits/9418fb52cde7872a7687ef20251e5633912cbd9c

frontend/src/common/updatecheck.ts
infeo: See https://github.com/cryptomator/hub/pull/121#discussion_r945896977

tobihagemann (approved):

PR #132, opened 2022-08-19
Update user and related data on login

### LINKED ISSUE #110: Update user data on login
Users and groups are synced in regular intervals with Keycloak. If a user is removed within this interval from the group, in Hub they are still part of the group.

To mitigate such security relevant issues, the user metadata should be synced on every login.

infeo:
Closes #110 

Thie PR implements the `UserResource::syncMe` method. The single-user-sync is implemented in the `RemoteUserPuller` due to already existing code. All group data is synchronized, because the user can be part of a new, not synced group. The synchronization itself is transactional annotated.

The syncMe method is called, when the user is freshly logged in. This is noted by adding the query parameter `login=true` to the redirect-url after successful keycloak login

Remark: There are two pending TODOS, would be wonderful if those can be fixed. Additionally, the `RemoteUserPuller::syncSingleUser(...)` lacks testing.

### DIFF
--- a/backend/src/main/java/org/cryptomator/hub/KeycloakRemoteUserProvider.java
+++ b/backend/src/main/java/org/cryptomator/hub/KeycloakRemoteUserProvider.java
@@ -28,6 +28,15 @@ public class KeycloakRemoteUserProvider implements RemoteUserProvider {
 	SyncerConfig syncerConfig;
 
 	@Override
+	public User user(String id) {
+		try (Keycloak keycloak = Keycloak.getInstance(syncerConfig.getKeycloakUrl(), syncerConfig.getKeycloakRealm(), syncerConfig.getUsername(), syncerConfig.getPassword(), syncerConfig.getKeycloakClientId())) {
+			return Optional.ofNullable(keycloak.realm(syncerConfig.getKeycloakRealm()).users().get(id).toRepresentation()) //
+					.map(this::mapToUser) //
+					.orElse(null);
+		}
+	}
+
+	@Override
 	public List<User> users() {
 		try (Keycloak keycloak = Keycloak.getInstance(syncerConfig.getKeycloakUrl(), syncerConfig.getKeycloakRealm(), syncerConfig.getUsername(), syncerConfig.getPassword(), syncerConfig.getKeycloakClientId())) {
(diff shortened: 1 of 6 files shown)


SailReal (changes requested):
In general, we need another algorithm to query a specific user as this method is called on each user login. Because as you note in

> All group data is synchronized, because the user can be part of a new, not synced group.

that means one request for all groups, one request per group to get the group details, up to `n` requests per group to get all members, `n` is `members per group` / 5_000. This means a lot of traffic against Keycloak, which very likely results quiet early in a DOS.

But I'm not quite sure yet how the algorithm should actually look like, the problem is the new groups (not already known to Hub), otherwise we could just do
1. load user with its groups from Keycloak (by adding `keycloak.realm(realm).users().get(id).groups()` which doesn't load the members of each group so only one request)
2. load user from db
3. delete user from the groups where they are not a member anymore
4. delete user from the groups where they are new, except those that are not yet known to Hub
5. update all other parameters

However, now we have the problem that the user's persistent groups are not complete. If we want to accept this for performance reasons we would be done now (variant a). A later run of the syncer would sync the missing groups. Otherwise we could now synchronize the groups that are not yet known to Hub (variant b), which could still be a lot of traffic depending on the setup, but that would be done once and skipped at the next login.

I'm not sure right now if I prefer variant a or b, I'm still thinking about it, maybe there is also another possibility...

P.S.: Because there are several requests against keycloak we can't do this completely in a `@Transactional` scope. So first query everything from keycloak (e.g. own method) and then in another do the queries against the database in the `@Transactional` scope. Otherwise we run into timeouts.

overheadhunter (changes requested):
I believe we *should not* sync groups.

Correct me, if I'm wrong, but groups are only relevant when accessing a vault or when granting access to a vault. They are not required for logging in.

In other words: While not _optimal_, it is _acceptable_ to have a short period of time between the implicit sync of the user data and the next "full sync" in which the user does not yet gain access to a vault due to (effective) group membership.

Logging in can be considered a frequent operation and should not trigger any expensive operation.

SailReal:
We decided to update the scope of this issue to only update/create the user without touching group memberships, therefore closing this PR an created https://github.com/cryptomator/hub/pull/135

Issue #134, opened 2022-08-30
Update user membership on login


SailReal:
### Please agree to the following

- [X] I have searched [existing issues](https://github.com/cryptomator/hub/issues?q=) for duplicates
- [X] I agree to follow this project's [Code of Conduct](https://github.com/cryptomator/hub/blob/develop/.github/CODE_OF_CONDUCT.md)

### Summary

Update user membership on login

### Motivation

Users and groups are synced in regular intervals with Keycloak. If a user is removed within this interval from the group, in Hub they are still part of the group.

To mitigate such security relevant issues, the user group membership should be synced on every login.

### Anything else?

The problem currently is, that the claim of the Keycloak group mapper can only contain the path and or name of the group, not the ID. Updating group memberships based on the path isn't a good idea. We need to wait until group IDs can be added to the claim.

PR #135, opened 2022-08-30
Updating user and associated data at login except group memberships

### LINKED ISSUE #110: Update user data on login

SailReal:
We decided to keep the implementation of #110 as simple as possible. All information to update the user is already available in the access token (JWT) signed by Kecloak so we can use it to update the logged-in user. Group membership can be added to the claims only by name or path not the ID so isn't useful to update membership. Created #134 for this part.

Fixes #110 Closes #132

### DIFF
--- a/frontend/src/router/index.ts
+++ b/frontend/src/router/index.ts
@@ -121,7 +121,7 @@ router.beforeEach((to, from, next) => {
     authPromise.then(async auth => {
       if (auth.isAuthenticated()) {
         await backend.users.syncMe();
-        delete to.query.sync_me;
+        delete to.query.sync_me; // remove sync_me query parameter to avoid endless recursion
         next({ path: to.path, query: to.query, params: to.params, replace: true });
       } else {
         next();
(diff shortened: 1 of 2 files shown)


frontend/src/router/index.ts
```diff
@@ -117,12 +117,16 @@ router.beforeEach((to, from, next) => {
 
 // SECOND update user data (requires auth)
 router.beforeEach((to, from, next) => {
-  if ('login' in to.query) {
+  if ('sync_me' in to.query) {
     authPromise.then(async auth => {
       if (auth.isAuthenticated()) {
         await backend.users.syncMe();
+        delete to.query.sync_me;
```
overheadhunter, line 124: is this to avoid endless recursion? please add a comment explaining it

overheadhunter (commented):
Looks good, just wondering what the removal of the query param is for

infeo (approved):

frontend/src/router/index.ts
SailReal: Exactly, fixed in https://github.com/cryptomator/hub/pull/135/commits/36007654d94be552132380fcf19c6bb563cdef9d

overheadhunter (approved):

PR #138, opened 2022-09-06
Add link to keycloak instance in frontend for admin (Fixes #107)

### LINKED ISSUE #107: Add link to keycloak instance in frontend for admin
As admin I want to access the keycloak instance from within the hub frontend.

Reasoning: Currently, Keycloak is needed to setup users/groups and ID Providers. To quickly check the settings and always access the associated keycloak instance, a clickabel link/button which redirects to keycloak would be nice.

SailReal:
Screenshot from 2022-09-06 12-55-21

### DIFF
--- a/frontend/src/components/AdminSettings.vue
+++ b/frontend/src/components/AdminSettings.vue
@@ -226,6 +226,7 @@ const betaUpdateExists = computed(() => {
 });
 
 onMounted(async () => {
+  keycloakURL.value = config.get().keycloakUrl;
   if (props.token) {
     await setToken(props.token);
   }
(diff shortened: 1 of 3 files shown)


infeo:
@tobihagemann After updating heroicons 2.0.0 i suggest to use https://github.com/tailwindlabs/heroicons/blob/master/src/24/solid/arrow-top-right-on-square.svg. What do you think?

frontend/src/components/AdminSettings.vue
```diff
@@ -241,6 +246,7 @@ async function fetchData() {
     let versionAvailable = versionDto.then(versionDto => updateChecker.get(versionDto.hubVersion));
     let adminDto = backend.billing.get();
     admin.value = await adminDto;
+    keycloakURL.value = config.get().keycloakUrl;
```
infeo, line 249: I suggest to move this assignment into the `onMounted(...)` method, since the config object is already present. At that point in time, there is no data fetching for it anymore

infeo (changes requested):

frontend/src/components/AdminSettings.vue
SailReal: Applied in https://github.com/cryptomator/hub/pull/138/commits/c54fb53d1473625d7ac89a8d58e51697c7cf62b3

infeo (approved):

PR #145, opened 2022-09-07
Search for new vault members via hub and not via Keycloak, fixes #139

### LINKED ISSUE #139: Adding a newly created user/group to vault members may result in a "Resource not found" error

### Summary

Don't search directly against keycloak but the Hub databse or sync the freshly created user/group when selected and not found

### System Setup

```markdown
- Hub: 1.0.0-beta3
```

### Steps to Reproduce

1. Create a new user `user3000`
2. Directly try to add user `user3000` as member to a vault before the syncer runs

### Expected Behavior

User "user3000" is added to the vault members or is not included in the list at all

### Actual Behavior

Screenshot from 2022-09-06 16-07-33

### Reproducibility

Always

### Relevant Log Output

_No response_

### Anything else?

The problem is that the search for users/groups runs against Keycloak. If a user/group was freshly created and not yet synchronized to hub, the error occurs.

SailReal:


backend/src/main/java/org/cryptomator/hub/api/AuthorityResource.java
```diff
@@ -0,0 +1,38 @@
...
+	@NoCache
+	@Operation(summary = "search authority")
+	public List<AuthorityDto> search(@QueryParam("query") String query) {
+		return Authority.byName(query).map(authority -> {
+			if (authority instanceof User user) {
+				return new UserDto(authority.id, authority.name, user.pictureUrl, user.email, null);
+			} else if (authority instanceof Group) {
+				return new GroupDto(authority.id, authority.name);
+			}
+			throw new IllegalStateException("authority is not of type user or group");
```
overheadhunter, line 34: Can Authority be a sealed class, so the compiler knows that those two ifs are exhaustive?

Alternatively, I would prefer to put this throw statement into an else block for easier readability (you can see the branches are mutually exclusive without reading what happens inside the ifs)

frontend/src/components/VaultDetails.vue
```diff
@@ -226,8 +226,7 @@ function permissionGranted() {
 }
 
 async function searchAuthority(query: string): Promise<AuthorityDto[]> {
-  return (await Promise.all([backend.users.search(query), backend.groups.search(query)]))
-    .flat()
+  return (await backend.authorities.search(query))
```
overheadhunter, line 229: Compared with the previous implementation this looses type information. Not a big deal in js, but still. Isn't the UI switching on the type in order to show different icons?

backend/src/main/java/org/cryptomator/hub/api/AuthorityResource.java
SailReal: I get a runtime using a sealed class `Caused by: java.lang.IncompatibleClassChangeError: class org.cryptomator.hub.entities.Authority$HibernateProxy$qGonujMw cannot inherit from sealed class org.cryptomator.hub.entities.Authority` so I implemented the other variant in https://github.com/cryptomator/hub/pull/145/commits/af5a66aa5d2078716e3c79112eeec2838f4161e1

frontend/src/components/VaultDetails.vue
SailReal: Fixed in https://github.com/cryptomator/hub/pull/145/commits/b81b2fcbcabcac78104bd34dfa7428beb66b87d6 and already discussed it with @tobihagemann

SailReal: With https://github.com/cryptomator/hub/pull/145/commits/f2ebc7d5f0ae3ad2765701dd560f0daa3e2c2d52 all DTOs that are still classes are now instantiated with a type guard after validating the type, all others are types. In a later step the plan is to provide the other DTOs also with type-guards and to map these corresponding responses from the server as well (analogous to User/GroupDto).

overheadhunter (approved):

PR #146, opened 2022-09-09
configurable base path

### LINKED ISSUE #144: Allow configuration of a base path

### Summary

Allow running Hub under arbitrary base URLs

### Motivation

I don't want to create a new domain for Hub, I want to serve it under a subresource of my existing domain.

To some extent, this can be done using a reverse proxy, however this only affects requests to the backend. The JS frontend deploys its own router, which is evaluated in the browser and thus not affected by my server-side proxy. Therefore a configuration value is required, which is available to both the Quarkus backend and Vue frontend.

### Considered Alternatives

Let the Vue router detect the current base path by reading `window.location`. However, this may cause _any_ path to be considered the base path, leading to infinite recursion in the router. The base path needs to be constant.

### Anything else?

This Traefik middleware has been used for testing whether one could reach the backend under arbitrary paths:

```yml
apiVersion: traefik.containo.us/v1alpha1
kind: Middleware
metadata:
  name: strip-prefix
spec:
  stripPrefix:
    prefixes:
      - /myCustomBasePath
---
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: ingress
  annotations:
    traefik.ingress.kubernetes.io/router.middlewares: strip-prefix@kubernetescrd
spec:
  rules:
    - host: example.com
      http:
        paths:
          - path: /myCustomBasePath
            pathType: Prefix
            backend:
              service:
                name: cryptomator-hub-service
                port:
                  number: 8080
```

overheadhunter:
This fixes #144 by introducing a new property `hub.public-root-path`, which can be specified at deploy time.


frontend/src/router/index.ts
```diff
@@ -80,35 +85,36 @@ const routes: RouteRecordRaw[] = [
...
   },
   {
-    path: '/unlock-error',
+    path: '/app/unlock-error',
     component: UnlockError,
     meta: { skipAuth: true }
   },
   {
-    path: '/:catchAll(.*)', //necessary due to using history mode in router
+    path: '/app/:pathMatch(.*)', //necessary due to using history mode in router
```
SailReal, line 98:
```suggestion
    path: '/:pathMatch(.*)', //necessary due to using history mode in router
```

Should we switch it to `/` so that also something like `http://localhost:3000/foo` shows the 404 error page? In the current implementation it throws the following error 

```
[Vue Router warn]: No match found for location with path "/foo" vue-router.mjs:35:17
Uncaught (in promise) Error: No match for
 {"query":{"sync_me":null},"params":{}}
while being at
{"fullPath":"/foo","path":"/foo","query":{},"hash":"","params":{},"matched":[],"meta":{},"href":"/foo"}
    createRouterError vue-router.mjs:799
    resolve vue-router.mjs:1444
    resolve vue-router.mjs:3024
    <anonymous> index.ts:117
    promise callback* index.ts:115
    guardToPromiseFn vue-router.mjs:1979
    guardToPromiseFn vue-router.mjs:1953
    runGuardQueue vue-router.mjs:3536
    promise callback*runGuardQueue/< vue-router.mjs:3536
    runGuardQueue vue-router.mjs:3536
    navigate vue-router.mjs:3218
    promise callback*navigate vue-router.mjs:3211
    pushWithRedirect vue-router.mjs:3143
    push vue-router.mjs:3078
    install vue-router.mjs:3499
    use runtime-core.esm-bundler.js:4327
    <anonymous> main.ts:19
vue-router.mjs:799:22
```

SailReal (changes requested):
The need for this functionality I totally understand but phew it doesn't make things easier :sweat_smile:

Except for those two little things, it looks good to me. Have also tested it locally, but not deployed it somewhere yet.

frontend/src/router/index.ts
overheadhunter: Tried, but behaves incorrectly as it collides with "real" 404 responses from the server. The Vue router should stay within its bounds.

backend/src/main/resources/application.properties
```diff
@@ -3,6 +3,10 @@
 # * or env vars `QUARKUS_HTTP_PORT=8080`
 # see: https://quarkus.io/guides/config-reference#configuration-sources
 
+# The
```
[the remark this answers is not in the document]
overheadhunter: fixed in b234974

SailReal (approved):

PR #150, opened 2022-09-23
Generate identicons if no profile picture is set

overheadhunter:
This replaces the hard coded "user" and "group" fallback images by identicons generated from the user/group id.

Different color configurations is used for users and groups to make the latter more distinctive in mixed lists. Fine tuning wouldn't hurt though (@tobihagemann 😉)

Screenshot 2022-09-23 at 13 08 06


overheadhunter:
In fact I believe it would be best to add a few more hues, so we get more distinctive images. Using high saturation for users and low for groups.

tobihagemann (approved):

PR #151, opened 2022-09-23
Frontend: Resolve paths relative to `<base href="..."/>`

overheadhunter:
This change replaces the former `/%hub.public-root-path%/` pattern that got replaced in _all_ static files by the backend, including .js, .css, etc.

Instead we will now only change .html by replacing the url in `<base href="/"/>` with the configured value from `hub.public-root-path`.

Not changing any other resource is not only a simplification but also a preparation for SRI. While SRI is not strictly necessary when loading subresources from the same origin (note that our CSP disallows any other origins), it doesn't hurt and may help future-proofing this system in case we need to relax the CSP.


overheadhunter:
Btw the reason why I didn't enable SRI yet (see 9d90367), is because vite doesn't support it natively and the only plugin I got working reports incorrect hashes for the js file...

frontend/index.html
```diff
@@ -1,8 +1,9 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
+    <base href="/"/>
     <meta charset="UTF-8" />
-    <link rel="icon" href="/favicon.ico" />
+    <link rel="icon" href="favicon.ico" />
```
tobihagemann, line 6: Not sure if this change was really necessary. Is it because it's more precise since the base tag only applies to relative URLs? Just wanted to note that `/favicon.ico` would've also worked (as far as I understand, only because the base in the vite config has been set to `./`).

tobihagemann (approved):
Seems to work just fine. Added one comment (not a request for change, just some feedback) and I've also noticed the issue #152.

frontend/index.html
overheadhunter: I still need to understand what Vite does here. I assumed it would only correct absolute paths when defining a base, leaving relative urls unchanged. I was wrong.
What I am really looking for is a way to use relative urls the way I intended them to be without Vite trying to be "intelligent".

overheadhunter: undone in d4d3ee6

PR #154, opened 2022-09-26
Feature: REST input parameter validation

### LINKED ISSUE #149: Validate REST input parameters

### Summary

Validate REST input parameters where necessary

### Motivation

To prevent execution of foreign/malicous code embedded into input parameters.

### Considered Alternatives

_No response_

### Anything else?

Quarkus provides a guide for REST parameter validation: https://quarkus.io/guides/validation

infeo:
Closes #149 

This PR adds validation of REST _input_ paramters. Output is not validated.
Since certain patterns with same set of restrictions are common (e.g. id for an object), i created `ValidationUtil.java` to bundle such patterns.

Key material fields (i.e., for authentification or encryption) have nearly not restrictions, since i didn't want to assume encoding, but can be changed.


backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
```diff
@@ -302,10 +307,12 @@ public Response create(@PathParam("vaultId") String vaultId, VaultDto vaultDto)
 		}
 	}
 
-	public record VaultDto(@JsonProperty("id") String id, @JsonProperty("name") String name, @JsonProperty("description") String description,
+	public record VaultDto(@JsonProperty("id") @NotBlank @Pattern(regexp = ValidationUtil.ID_PATTERN) String id,
+						   @JsonProperty("name") @NotBlank @Pattern(regexp = ValidationUtil.NAME_PATTERN) String name,
+						   @JsonProperty("description") @Pattern(regexp = "(?U)[-\\p{Alnum}\\h_.~!%&'()*+;=:/?#@]*") String description,
```
overheadhunter, line 312: pure madness :D this is too much complexity, which may introduce errors. if we want to allow _some_ legal content,  such as (reading your tests) URLs, please use a different approach:

Create a new annotation class `SafeText`, which itself is annotated with `@Constraint(validatedBy = SafeTextValidator.class)`.  Then add org.owasp.esapi to our dependencies and create the `class SafeTextValidator implements ConstraintValidator<...>` which calls esapi for input validation.

We should then also use DOMPurifier on the frontend, but this is a different PR.

overheadhunter (changes requested):

backend/src/main/java/org/cryptomator/hub/api/ValidationUtil.java
```diff
@@ -0,0 +1,11 @@
+package org.cryptomator.hub.api;
+
+public final class ValidationUtil {
+
+	public static final String ID_PATTERN = "[-\\w]+";
+	public static final String JWX_PATTERN = "[-=.\\w]+";
```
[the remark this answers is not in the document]
infeo: Base64Url uses `=` for padding, if needed (see https://datatracker.ietf.org/doc/html/rfc4648#page-7).

Additional, if `.` appears in a group, it loses its special meaning of a match all expression, see https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html?is-external=true, section "Character Classes".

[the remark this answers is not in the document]
overheadhunter: > note that JWX use base64url (_-), not base64 (/+=).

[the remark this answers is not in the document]
overheadhunter: > Additional, if `.` appears in a group

yes but when you use `.` in a regex, you match *any char*, not `.`. What you want to use is`\.`

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
```diff
@@ -302,10 +307,12 @@ public Response create(@PathParam("vaultId") String vaultId, VaultDto vaultDto)
...
 						   @JsonProperty("creationTime") @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd'T'HH:mm:ss.SSSX") Timestamp creationTime,
-						   @JsonProperty("masterkey") String masterkey, @JsonProperty("iterations") String iterations, @JsonProperty("salt") String salt,
-						   @JsonProperty("authPublicKey") String authPublicKey, @JsonProperty("authPrivateKey") String authPrivateKey
+						   @JsonProperty("masterkey") @NotBlank String masterkey, @JsonProperty("iterations") @NotBlank @Pattern(regexp = "\\d+") String iterations, @JsonProperty("salt") @NotNull String salt,
+						   @JsonProperty("authPublicKey") @NotBlank String authPublicKey, @JsonProperty("authPrivateKey") @NotBlank String authPrivateKey
```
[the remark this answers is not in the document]
overheadhunter: in fact it seems like masterkey is base64url and admin auth keypair is base64 🤔

backend/src/main/java/org/cryptomator/hub/api/ValidationUtil.java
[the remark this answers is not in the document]
infeo: In a regular expresion, if `.` is used in a character class, it _does not_ macht any char. (since it loses its special meaning inside of it). It only matches the dot character itself.
```
jshell> Pattern.matches("[a.]+", "bb")
$1 ==> false

jshell> Pattern.matches("[a.]+", "a.a")
$3 ==> true
```

infeo:
I did an overhaul of the approach:
1. Be as specific as possible
2. Create own annotations describing the pattern (not the target)

I have to add tests, but i think this apporach increases the readability and reduces maintenance

backend/src/main/java/org/cryptomator/hub/api/ValidationUtil.java
```diff
@@ -0,0 +1,11 @@
+package org.cryptomator.hub.api;
+
+public final class ValidationUtil {
```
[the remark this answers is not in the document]
infeo: outdated

backend/src/main/java/org/cryptomator/hub/api/ValidationUtil.java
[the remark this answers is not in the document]
infeo: outdated

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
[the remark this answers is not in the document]
infeo: After merging https://github.com/cryptomator/hub/pull/154/commits/291f89ec2b4c8a727ff06e69c397a854ac8d9871, masterkey, salt, pubKey and privKey are base64 encoded.

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
```diff
@@ -307,12 +312,12 @@ public Response create(@PathParam("vaultId") @Pattern(regexp = ValidationUtil.ID
 		}
 	}
 
-	public record VaultDto(@JsonProperty("id") @NotBlank @Pattern(regexp = ValidationUtil.ID_PATTERN) String id,
-						   @JsonProperty("name") @NotBlank @Pattern(regexp = ValidationUtil.NAME_PATTERN) String name,
-						   @JsonProperty("description") @Pattern(regexp = "(?U)[-\\p{Alnum}\\h_.~!%&'()*+;=:/?#@]*") String description,
+	public record VaultDto(@JsonProperty("id") @IsUUID String id,
+						   @JsonProperty("name") @AlmostSafeText String name,
+						   @JsonProperty("description") @Pattern(regexp = "^[&*<>\"]") String description,
```
overheadhunter, line 317: `^` at this position means "start of string/line", you need to put it inside the `[^...]` .

backend/src/main/java/org/cryptomator/hub/api/ValidationPatterns.java
```diff
@@ -1,11 +1,10 @@
 package org.cryptomator.hub.api;
 
-public final class ValidationUtil {
+public final class ValidationPatterns {
 
 	public static final String ID_PATTERN = "[-\\w]+";
```
overheadhunter, line 5: is this still used?

backend/src/main/java/org/cryptomator/hub/validation/SafeText.java
```diff
@@ -0,0 +1,32 @@
...
+import static java.lang.annotation.ElementType.TYPE_USE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * TODO: doc doc doc
+ */
+@Pattern(regexp = "[-_.a-zA-Z0-9]*")
+@Target({METHOD, FIELD, ANNOTATION_TYPE, TYPE_USE})
+@Retention(RUNTIME)
+@Constraint(validatedBy = {})
```
overheadhunter, line 22: I think you can remove these, as they're inherited from `@Pattern`

backend/src/main/java/org/cryptomator/hub/api/ValidationPatterns.java
```diff
@@ -1,11 +1,10 @@
...
-	public static final String JWX_PATTERN = "[-=.\\w]+";
 	public static final String NAME_PATTERN = "(?U)[-_\\p{Alpha}\\p{Digit}\\x20]+";
```
overheadhunter, line 6: is this still used?

backend/src/main/java/org/cryptomator/hub/validation/AlmostSafeText.java
```diff
@@ -0,0 +1,29 @@
...
+import static java.lang.annotation.ElementType.TYPE_USE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Pattern(regexp = "(?Ui)[-_\\w\\s]+")
+@NotBlank
+@Target({METHOD, FIELD, ANNOTATION_TYPE, TYPE_USE})
+@Retention(RUNTIME)
+@Constraint(validatedBy = {})
+@Documented
+public @interface AlmostSafeText {
```
overheadhunter, line 23: This is a terrible name. Looking at the usage, I would suggest renaming this to something like "UserFacingText" and renaming SafeText to ValidID

overheadhunter (changes requested):
I believe there is still some unused code. Furthermore instead of `@IsXYZ` I would prefer `@ValidXYZ` to better transport the purpose of these annotations.

backend/src/main/java/org/cryptomator/hub/validation/ValidPseudoBase64.java
```diff
@@ -0,0 +1,28 @@
...
+import static java.lang.annotation.ElementType.*;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Pattern(regexp = "[+/A-Za-z0-9]+=*")
+@NotNull
+@Target({METHOD, FIELD, ANNOTATION_TYPE, TYPE_USE, PARAMETER})
+@Retention(RUNTIME)
+@Constraint(validatedBy = {})
+@Documented
+public @interface ValidPseudoBase64 {
```
overheadhunter, line 21: I guess by `Pseudo` you're referring to the fact that this check doesn't check whether decoding is actually possible?

backend/src/main/java/org/cryptomator/hub/validation/ValidName.java
```diff
@@ -0,0 +1,26 @@
...
+import static java.lang.annotation.ElementType.*;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Pattern(regexp = "(?Ui)[-_\\w\\s]+")
+@NotBlank
+@Target({METHOD, FIELD, ANNOTATION_TYPE, TYPE_USE, PARAMETER})
+@Retention(RUNTIME)
+@Constraint(validatedBy = {})
+@Documented
+public @interface ValidName {
```
overheadhunter, line 20: Imho this is still too specific. I would like you to only create custom annotations for "types" such as JWEs, UUIDs, etc, not for "meanings" such as names, descriptions, comments, ... In the latter cases, just apply the `@Pattern` annotation directly where needed.

(Also, it doesn't match `Nick O’Neill`.)

backend/src/main/java/org/cryptomator/hub/validation/ValidUUID.java
```diff
@@ -0,0 +1,27 @@
...
+import static java.lang.annotation.ElementType.*;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Pattern(regexp = "[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}")
+@NotNull
+@Target({METHOD, FIELD, ANNOTATION_TYPE, TYPE_USE, PARAMETER})
+@Retention(RUNTIME)
+@Constraint(validatedBy = {})
+@Documented
+public @interface ValidUUID {
```
overheadhunter, line 21: I just noticed that we use IDs in our test data that are *not* UUIDs. Maybe we should drop this validation again? `ValidID` is sufficient to protect against malformed data.

overheadhunter (changes requested):

backend/src/main/java/org/cryptomator/hub/api/ValidationPatterns.java
overheadhunter: @infeo is it?

infeo: ow, sh*te, completely forgot. Will fix it.

overheadhunter: wait until you see my other comment regarding the removal of `@ValidName`, though 😜

infeo: Done in [002a11c](https://github.com/cryptomator/hub/pull/154/commits/002a11ccf5b2cdcc5b6be018d76304bfe167ab25)

backend/src/test/java/org/cryptomator/hub/api/ValidationUtilTest.java
```diff
@@ -0,0 +1,43 @@
...
+	@DisplayName("Test id pattern with matches")
+	@ValueSource(strings = {"asd-asd", "029foobar-b0z"})
+	public void testIdPatternValid(String matches) {
+		var result = Pattern.matches(ValidationUtil.ID_PATTERN, matches);
+		Assertions.assertTrue(result);
+	}
+
+	@ParameterizedTest
+	@DisplayName("Test id pattern with no-matches")
+	@ValueSource(strings = {"foo\u5207", "asd§", "asd$", "%&asd", "as02nmf-laksdj.", "foo/\\bar", "+foobarbaz#-\"", "<bar>"})
```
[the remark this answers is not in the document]
infeo: Done in https://github.com/cryptomator/hub/pull/154/commits/1461d1528b523955a86e7ecaac1261c5fe6d369b

backend/src/main/java/org/cryptomator/hub/validation/ValidUUID.java
infeo: Done in https://github.com/cryptomator/hub/pull/154/commits/76ec596ab1308f5556c0bf73fbc9ded0ff876341

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
infeo: Fixed in https://github.com/cryptomator/hub/pull/154/commits/7960c3fa1ea1cdad4ac7d2172a68454b398227f8

backend/src/main/java/org/cryptomator/hub/validation/ValidPseudoBase64.java
overheadhunter: Then maybe rename it to `OnlyBase64Chars`

backend/src/main/java/org/cryptomator/hub/validation/ValidName.java
infeo: See https://github.com/cryptomator/hub/pull/154/commits/e1230648ffa6ddcdf4043565e1f1ee2f091249e4

backend/src/main/java/org/cryptomator/hub/validation/ValidPseudoBase64.java
infeo: Fixed in https://github.com/cryptomator/hub/pull/154/commits/90d2dbc60add1185cbfa1ee4be7a2acba01fd50f

overheadhunter (approved):

PR #156, opened 2022-09-27
Point link in Hub admin tab to Keycloak Hub realm not Master realm

### LINKED ISSUE #107: Add link to keycloak instance in frontend for admin

SailReal:
Fixes #107

### DIFF
--- a/frontend/src/components/AdminSettings.vue
+++ b/frontend/src/components/AdminSettings.vue
@@ -226,7 +226,8 @@ const betaUpdateExists = computed(() => {
 });
 
 onMounted(async () => {
-  keycloakAdminRealmURL.value = config.get().keycloakAdminRealmUrl;
+  let cfg = config.get();
+  keycloakAdminRealmURL.value = `${cfg.keycloakUrl}/admin/${cfg.keycloakRealm}/console`;
   if (props.token) {
     await setToken(props.token);
   }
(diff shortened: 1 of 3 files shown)


frontend/src/components/AdminSettings.vue
```diff
@@ -226,7 +226,7 @@ const betaUpdateExists = computed(() => {
 });
 
 onMounted(async () => {
-  keycloakURL.value = config.get().keycloakUrl;
+  keycloakAdminRealmURL.value = config.get().keycloakAdminRealmUrl;
```
overheadhunter, line 229: Since we already have both the Keycloak base url as well as the realm id, I believe we can do the concatenation part in javascript, keeping the DTO small.

```suggestion
  let cfg = config.get();
  keycloakAdminRealmURL.value = `${cfg.keycloakUrl}/admin/${cfg.keycloakRealm}/console`;
```

overheadhunter (commented):
I would suggest moving the string concatenation to js, which keeps both the DTOs and the diff small.

frontend/src/components/AdminSettings.vue
SailReal: I also thought about it but came to the conclusion that it is rather the task of the backend to provide such URLs but if you want, I'll change it.

overheadhunter: I think what I dislike about it coming from the backend is that it suggests this url could be configured separately. When in reality it is not really a standalone configuration but rather a derived property.

overheadhunter (approved):

PR #158, opened 2022-09-28
List all vaults as admin and search in vault list

### LINKED ISSUE #106: List all vaults as admin
Since the hub administrator manages the billing, he/she should also have an overview over all vaults. ~This does not mean, that he/she can edit or remove them, but listing them and seeing the members should be possible.~

SailReal:
Fixes #106

https://user-images.githubusercontent.com/1786772/192776811-93c89145-f9cf-4b56-908f-9cc0b7e57162.mp4

### DIFF
--- a/backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
+++ b/backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
@@ -2,6 +2,7 @@ package org.cryptomator.hub.api;
 
 import com.fasterxml.jackson.annotation.JsonFormat;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import io.quarkus.security.identity.SecurityIdentity;
 import org.cryptomator.hub.entities.AccessToken;
 import org.cryptomator.hub.entities.Device;
 import org.cryptomator.hub.entities.EffectiveGroupMembership;
@@ -46,6 +47,9 @@ public class VaultResource {
 	JsonWebToken jwt;
 
 	@Inject
+	SecurityIdentity identity;
+
+	@Inject
 	LicenseHolder license;
 
 	@GET
@@ -58,6 +62,16 @@ public class VaultResource {
 		var currentUserId = jwt.getSubject();
 		var resultStream = Vault.findAccessibleOrOwnedByUser(currentUserId);
 		return resultStream.map(VaultDto::fromEntity).toList();
+	}
+
+	@GET
+	@Path("/all")
+	@RolesAllowed("admin")
+	@Produces(MediaType.APPLICATION_JSON)
+	@Transactional
+	@Operation(summary = "list all accessible vaults", description = "list all vaults in the system")
+	public List<VaultDto> getAllVaults() {
+		return Vault.findAll().<Vault>stream().map(VaultDto::fromEntity).toList();
 	}
 
 	@GET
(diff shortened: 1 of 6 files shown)


frontend/src/i18n/en-US.json
```diff
@@ -108,5 +108,12 @@
   "vaultList.title": "Vaults",
   "vaultList.empty.title": "No vaults",
   "vaultList.empty.description": "Get started by creating a new vault.",
-  "vaultList.createVault": "Create Vault"
+  "vaultList.createVault": "Create Vault",
+  "vaultList.filter": "Filter",
+  "vaultList.filter.entry.allVaults": "All vaults",
+  "vaultList.filter.entry.ownedByMe": "Owned by me",
```
overheadhunter, line 114: Shouldn't this rather be something like this?

```suggestion
  "vaultList.filter.entry.accessibleByMe": "Vaults I have access to",
```

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
```diff
@@ -60,6 +64,16 @@ public List<VaultDto> getSharedOrOwned() {
 		return resultStream.map(VaultDto::fromEntity).toList();
 	}
 
+	@GET
+	@Path("/all")
```
overheadhunter, line 68: Strictly speaking this is not a different resource, just a different filter. Is it possible to rather use `/api/vaults?all` without much hassle? If this requires adding a `@QueryParam` to the existing `getSharedOrOwned()` and requiring complicated branching depending on roles, I would prefer the current `/api/vaults/all` approach, though.

SailReal: Did some research and found no way to provide the correct response without branching or a request filter. At least the role must be checked manually because all vaults should only be called from the admin role.

frontend/src/i18n/en-US.json
SailReal: Hmm you're right but `Vaults I have access to` is quiet long
Screenshot from 2022-09-28 15-49-03

What do you think about `Accessible vaults`?
Screenshot from 2022-09-28 15-53-51

frontend/src/i18n/en-US.json
```diff
@@ -108,5 +108,12 @@
   "vaultList.title": "Vaults",
   "vaultList.empty.title": "No vaults",
   "vaultList.empty.description": "Get started by creating a new vault.",
-  "vaultList.createVault": "Create Vault"
+  "vaultList.createVault": "Create Vault",
+  "vaultList.filter": "Filter",
+  "vaultList.filter.entry.allVaults": "All vaults",
+  "vaultList.filter.entry.ownedByMe": "Accessible vaults",
```
tobihagemann, line 114:
```suggestion
  "vaultList.filter.entry.accessibleVaults": "Accessible vaults",
```

tobihagemann (changes requested):

tobihagemann (approved):

frontend/src/components/VaultList.vue
```diff
@@ -81,21 +116,45 @@ const onFetchError = ref<Error | null>();
...
   nextTick(() => vaultDetailsSlideOver.value?.show());
 }
+
+async function onAllVaultsClick() {
+  vaults.value = await backend.vaults.listAll();
+}
+
+async function onSharedOrOwnedVaultsClick() {
+  vaults.value = await backend.vaults.listSharedOrOwned();
+}
```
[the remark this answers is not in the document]
SailReal: Fixed in https://github.com/cryptomator/hub/pull/158/commits/b584fdd703ffdd2097c3929a8805e320c98496d9

PR #164, opened 2022-10-10
AdminJWTs are valid when they are issued in the future in defined leeway

SailReal:
Before the leeway calculation was in the client, now it is on the server side. Furthermore when the IAT was in the future, it failed before. Now the client provides only a IAT and we allow it to be in the future.

https://github.com/cryptomator/hub/blob/4aeaf04d19ad4841ff761defe92874c9711ff541/backend/src/main/java/org/cryptomator/hub/filters/VaultAdminOnlyFilterProvider.java#L78-L81

is required because if we use `.acceptIssuedAt(REQUEST_LEEWAY_IN_SECONDS)` the issue is valid until forever. That is why we need to validate it our self.

### DIFF
--- a/backend/src/main/java/org/cryptomator/hub/filters/VaultAdminOnlyFilterProvider.java
+++ b/backend/src/main/java/org/cryptomator/hub/filters/VaultAdminOnlyFilterProvider.java
@@ -21,6 +21,9 @@ import java.security.NoSuchAlgorithmException;
 import java.security.interfaces.ECPublicKey;
 import java.security.spec.InvalidKeySpecException;
 import java.security.spec.X509EncodedKeySpec;
+import java.sql.Date;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.Base64;
 
 @Provider
@@ -29,6 +32,7 @@ public class VaultAdminOnlyFilterProvider implements ContainerRequestFilter {
 
 	public static final String VAULT_ADMIN_AUTHORIZATION = "Cryptomator-Vault-Admin-Authorization";
 	static final String VAULT_ID = "vaultId";
+	static final int REQUEST_LEEWAY_IN_SECONDS = 15;
 
 	@Override
 	public void filter(ContainerRequestContext containerRequestContext) {
@@ -37,7 +41,7 @@ public class VaultAdminOnlyFilterProvider implements ContainerRequestFilter {
 		var unveridifedVaultId = getUnverifiedVaultId(vaultAdminAuthorizationJWT);
 		if (vaultIdQueryParameter.equals(unveridifedVaultId)) {
 			var vault = Vault.<Vault>findByIdOptional(unveridifedVaultId).orElseThrow(NotFoundException::new);
-			var algorithm = Algorithm.ECDSA384(decodePublicKey(vault.authenticationPublicKey), null);
+			var algorithm = Algorithm.ECDSA384(decodePublicKey(vault.authenticationPublicKey));
 			verify(buildVerifier(algorithm), vaultAdminAuthorizationJWT);
 		} else {
 			throw new VaultAdminValidationFailedException("Other vaultId provided");
@@ -48,11 +52,9 @@ public class VaultAdminOnlyFilterProvider implements ContainerRequestFilter {
 	void verify(JWTVerifier verifier, DecodedJWT vaultAdminAuthorizationJWT) {
 		try {
 			verifier.verify(vaultAdminAuthorizationJWT);
-		} catch (TokenExpiredException e) {
-			throw new VaultAdminTokenExpiredException("Token of VaultAdminAuthorizationJWT expired");
 		} catch (IncorrectClaimException e) {
-			if (e.getClaimName().equals(RegisteredClaims.ISSUED_AT) || e.getClaimName().equals(RegisteredClaims.NOT_BEFORE)) {
-				throw new VaultAdminTokenNotYetValidException("Token of VaultAdminAuthorizationJWT not yet valid");
+			if (e.getClaimName().equals(RegisteredClaims.ISSUED_AT)) {
+				throw new VaultAdminValidationFailedException("Token of VaultAdminAuthorizationJWT not provided or not yet or no longer valid");
 			} else {
 				throw new VaultAdminValidationFailedException("Incorrect claim exception");
 			}
@@ -66,12 +68,17 @@ public class VaultAdminOnlyFilterProvider implements ContainerRequestFilter {
 		return verification(algorithm).build();
 	}
 
+	private Verification verification(Algorithm algorithm) {
+		return verification(algorithm, Instant.now());
+	}
+
 	//visible for testing
-	Verification verification(Algorithm algorithm) {
+	Verification verification(Algorithm algorithm, Instant now) {
 		return JWT.require(algorithm) //
-				.withClaim(RegisteredClaims.ISSUED_AT, (claim, jwt) -> jwt.getIssuedAt() != null) //
-				.withClaim(RegisteredClaims.NOT_BEFORE, (claim, jwt) -> jwt.getNotBefore() != null) //
-				.withClaim(RegisteredClaims.EXPIRES_AT, (claim, jwt) -> jwt.getExpiresAt() != null);
+				.withClaim(RegisteredClaims.ISSUED_AT, (claim, jwt) -> jwt.getIssuedAt() != null //
+						&& !jwt.getIssuedAt().before(Date.from(now.minus(REQUEST_LEEWAY_IN_SECONDS, ChronoUnit.SECONDS))) //
+						&& !jwt.getIssuedAt().after(Date.from(now.plus(REQUEST_LEEWAY_IN_SECONDS, ChronoUnit.SECONDS)))) //
+				.ignoreIssuedAt();
 	}
 
 	//visible for testing
--- a/backend/src/test/java/org/cryptomator/hub/api/VaultResourceTest.java
+++ b/backend/src/test/java/org/cryptomator/hub/api/VaultResourceTest.java
@@ -1,5 +1,7 @@
 package org.cryptomator.hub.api;
 
+import com.auth0.jwt.JWT;
+import com.auth0.jwt.algorithms.Algorithm;
 import com.radcortez.flyway.test.annotation.DataSource;
 import com.radcortez.flyway.test.annotation.FlywayTest;
 import io.agroal.api.AgroalDataSource;
@@ -10,6 +12,7 @@ import io.quarkus.test.security.oidc.OidcSecurity;
 import io.restassured.RestAssured;
 import io.restassured.http.ContentType;
 import org.cryptomator.hub.filters.VaultAdminOnlyFilterProvider;
+import org.cryptomator.hub.filters.VaultAdminValidationFailedException;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
 import org.junit.jupiter.api.BeforeAll;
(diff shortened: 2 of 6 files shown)


backend/src/main/java/org/cryptomator/hub/filters/VaultAdminOnlyFilterProvider.java
```diff
@@ -66,12 +68,17 @@ JWTVerifier buildVerifier(Algorithm algorithm) {
...
 	//visible for testing
-	Verification verification(Algorithm algorithm) {
+	Verification verification(Algorithm algorithm, Instant now) {
 		return JWT.require(algorithm) //
-				.withClaim(RegisteredClaims.ISSUED_AT, (claim, jwt) -> jwt.getIssuedAt() != null) //
-				.withClaim(RegisteredClaims.NOT_BEFORE, (claim, jwt) -> jwt.getNotBefore() != null) //
-				.withClaim(RegisteredClaims.EXPIRES_AT, (claim, jwt) -> jwt.getExpiresAt() != null);
+				.withClaim(RegisteredClaims.ISSUED_AT, (claim, jwt) -> jwt.getIssuedAt() != null //
+						&& !jwt.getIssuedAt().before(Date.from(now.minus(REQUEST_LEEWAY_IN_SECONDS, ChronoUnit.SECONDS))) //
+						&& !jwt.getIssuedAt().after(Date.from(now.plus(REQUEST_LEEWAY_IN_SECONDS, ChronoUnit.SECONDS)))) //
```
overheadhunter, line 80: careful: these are `java.sql.Date`

backend/src/main/java/org/cryptomator/hub/filters/VaultAdminOnlyFilterProvider.java
```diff
@@ -48,11 +52,9 @@ public void filter(ContainerRequestContext containerRequestContext) {
 	void verify(JWTVerifier verifier, DecodedJWT vaultAdminAuthorizationJWT) {
 		try {
 			verifier.verify(vaultAdminAuthorizationJWT);
-		} catch (TokenExpiredException e) {
-			throw new VaultAdminTokenExpiredException("Token of VaultAdminAuthorizationJWT expired");
 		} catch (IncorrectClaimException e) {
-			if (e.getClaimName().equals(RegisteredClaims.ISSUED_AT) || e.getClaimName().equals(RegisteredClaims.NOT_BEFORE)) {
-				throw new VaultAdminTokenNotYetValidException("Token of VaultAdminAuthorizationJWT not yet valid");
+			if (e.getClaimName().equals(RegisteredClaims.ISSUED_AT)) {
+				throw new VaultAdminValidationFailedException("Token of VaultAdminAuthorizationJWT not provided or not yet or no longer valid");
```
overheadhunter, line 57: I really like the old exception better. We specifically test if the IAT claim is the culprit, so why should it be "VaultAdminAuthorizationJWT not provided"?

backend/src/test/java/org/cryptomator/hub/api/VaultResourceTest.java
```diff
@@ -42,21 +58,32 @@
...
+		var algorithmVault2 = Algorithm.ECDSA384((ECPrivateKey) getPrivateKey("MIG2AgEAMBAGByqGSM49AgEGBSuBBAAiBIGeMIGbAgEBBDCAHpFQ62QnGCEvYh/pE9QmR1C9aLcDItRbslbmhen/h1tt8AyMhskeenT+rAyyPhGhZANiAAQLW5ZJePZzMIPAxMtZXkEWbDF0zo9f2n4+T1h/2sh/fviblc/VTyrv10GEtIi5qiOy85Pf1RRw8lE5IPUWpgu553SteKigiKLUPeNpbqmYZUkWGh3MLfVzLmx85ii2vMU="));
+
+		vault1AdminJWT = JWT.create().withHeader(Map.of("vaultId", "vault1")).withIssuedAt(Instant.now()).sign(algorithmVault1);
+		vault2AdminJWT = JWT.create().withHeader(Map.of("vaultId", "vault2")).withIssuedAt(Instant.now()).sign(algorithmVault2);
+	}
+
+	private PrivateKey getPrivateKey(String keyBytes) throws NoSuchAlgorithmException, InvalidKeySpecException {
+		return KeyFactory.getInstance("EC").generatePrivate(new PKCS8EncodedKeySpec(Base64.getDecoder().decode(keyBytes)));
+	}
+
```
overheadhunter, line 86: 1. use a `@Before public void setup()` method rather than a constructor during tests.
2. it is a bad idea to use `Instant.now()` during tests, as the test is no longer deterministic. Use hard-coded time stamps and make the validator accept a specific date which should be considered as "now"

frontend/src/common/backend.ts
```diff
@@ -236,8 +235,7 @@ class VaultService {
 
   private async buildVaultAdminAuthorizationJWT(vaultId: string, vaultKeys: VaultKeys): Promise<string> {
     let vaultIdHeader: VaultIdHeader = { alg: 'ES384', b64: true, typ: 'JWT', vaultId: vaultId };
-    let nowInSeconds = this.secondsSinceEpoch();
-    let jwtPayload = { exp: nowInSeconds + vaultAdminAuthorizationJWTLeeway, nbf: nowInSeconds - vaultAdminAuthorizationJWTLeeway, iat: nowInSeconds };
+    let jwtPayload = { iat: this.secondsSinceEpoch() };
```
overheadhunter, line 238: What about the delta?

```js
let timeDiff = serverTime() - now(); // run this immediately after receiving the "config" response
// ... minutes later:
let iat = now() + config.getServerTimeDiff();
```

overheadhunter (changes requested):

backend/src/main/java/org/cryptomator/hub/filters/VaultAdminOnlyFilterProvider.java
SailReal: Fixed in https://github.com/cryptomator/hub/pull/164/commits/42e69c0ddc2006e41a41080a28bfd94e7029d58c

backend/src/main/java/org/cryptomator/hub/filters/VaultAdminOnlyFilterProvider.java
SailReal: The message wasn't correct, now it fits better: https://github.com/cryptomator/hub/pull/164/commits/42e69c0ddc2006e41a41080a28bfd94e7029d58c

Sure, if we want to throw a more specific exception, we need to check the cause again here... :thinking:

frontend/src/common/backend.ts
SailReal: Wanted to create a new PR with it, but here it is https://github.com/cryptomator/hub/pull/164/commits/d4cedb8dfcbb700c5bbc6d3a15a436b02f576d00

overheadhunter: instead of passing parameters around between vue components and backend.ts, can't the latter import `config` directly? It is a lazy singleton, after all.

SailReal: Correct :+1:, fixed in https://github.com/cryptomator/hub/pull/164/commits/6a0d3fedea5c73b8e75e887d03876c3d93483850

frontend/src/common/config.ts
```diff
@@ -14,12 +14,18 @@ const axios = AxiosStatic.create({
...
+  keycloakUrl: string;
+  keycloakClientId: string;
+  keycloakAuthEndpoint: string;
+  keycloakTokenEndpoint: string;
+  serverTime: string;
+};
 
 class ConfigWrapper {
   private data: ConfigDto;
+  private serverTimeDiff: number;
```
overheadhunter, line 28: instead of a private property and a public getter, use:

```suggestion
  readonly serverTimeDiff: number;
```

frontend/src/common/config.ts
```diff
@@ -32,12 +38,17 @@ class ConfigWrapper {
...
   }
 
   public get(): ConfigDto {
     return this.data;
   }
 
+  public getServerTimeDiff(): number {
+    return this.serverTimeDiff;
+  }
+
```
overheadhunter, line 51: see above

backend/src/main/java/org/cryptomator/hub/api/ConfigResource.java
```diff
@@ -68,7 +71,7 @@ String trimTrailingSlash(String str) {
 
 	public record ConfigDto(@JsonProperty("keycloakUrl") String keycloakUrl, @JsonProperty("keycloakRealm") String keycloakRealm,
 							@JsonProperty("keycloakClientId") String keycloakClientId, @JsonProperty("keycloakAuthEndpoint") String authEndpoint,
-							@JsonProperty("keycloakTokenEndpoint") String tokenEndpoint) {
+							@JsonProperty("keycloakTokenEndpoint") String tokenEndpoint, @JsonProperty("serverTime") @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd'T'HH:mm:ss.SSSX") Timestamp serverTime) {
```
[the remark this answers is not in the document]
SailReal: I just copy pasted it from https://github.com/cryptomator/hub/blob/develop/backend/src/main/java/org/cryptomator/hub/api/VaultResource.java#L306 but I agree, will fix it everywhere it is already done as well.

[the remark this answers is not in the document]
overheadhunter: If I am not mistaken, Jackson defaults to ISO 8601 strings when serializing Instants. So you might even be able to skip providing a date pattern.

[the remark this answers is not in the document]
SailReal: Tested it but using the default formatter results in the pattern `2022-10-12T08:04:26.125841097Z`. ISO_8601 states about this 

> There is no limit on the number of decimal places for the decimal fraction. However, the number of decimal places needs to be agreed to by the communicating parties. For example, in Microsoft SQL Server, the [precision](https://en.wikipedia.org/wiki/Significant_figures) of a decimal fraction is 3 for a DATETIME, i.e., "yyyy-mm-ddThh:mm:ss[.mmm]".[[27]](https://en.wikipedia.org/wiki/ISO_8601#cite_note-27)

`Date.parse()` of TypeScript obviously can not handle that much places and then refuses to create the `Date` object. So IMO we still need the formatter.

frontend/src/common/config.ts
```diff
@@ -32,12 +38,17 @@ class ConfigWrapper {
 
   private constructor(data: ConfigDto) {
     this.data = data;
+    this.serverTimeDiff = Math.floor(Date.parse(data.serverTime).valueOf() / 1000) - Math.floor(Date.now() / 1000);
```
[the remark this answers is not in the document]
SailReal: Implemented in https://github.com/cryptomator/hub/pull/164/commits/96a104ec834fc75db9695a822462c5ca6c0e2f0f and https://github.com/cryptomator/hub/pull/164/commits/e373fb727a461e684eae89efcd068326e0d67611

frontend/src/common/config.ts
SailReal: Implemented in https://github.com/cryptomator/hub/pull/164/commits/96a104ec834fc75db9695a822462c5ca6c0e2f0f and https://github.com/cryptomator/hub/pull/164/commits/e373fb727a461e684eae89efcd068326e0d67611

frontend/src/common/config.ts
SailReal: Implemented in https://github.com/cryptomator/hub/pull/164/commits/96a104ec834fc75db9695a822462c5ca6c0e2f0f and https://github.com/cryptomator/hub/pull/164/commits/e373fb727a461e684eae89efcd068326e0d67611

backend/src/main/java/org/cryptomator/hub/api/ConfigResource.java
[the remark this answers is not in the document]
overheadhunter: > `Date.parse()` of TypeScript obviously can not handle that much places and then refuses to create the `Date` object. So IMO we still need the formatter.

Fixed by truncating the instant to milliseconds 0c6b42d

overheadhunter (approved):

PR #170, opened 2022-10-18
Separate Hub and Cryptomator OIDC client to reduce the scopes of access

SailReal:
This will result in "breaking" changes, but existing installations can be updated. We will describe this at the time of release and the update of the setup wizard on the website will also be published then.

To updating an existing deployments the following steps are required:
1. remove in the `cryptomatorhub` client in `Valid redirect URIs` the following URL: `http://127.0.0.1`
2. create a new `cryptomator` (cryptomator is the id) `Open Id connect` client with only `Standard flow` enabled and set `Valid redirect URIs` to `http://127.0.0.1`
4. regenerate the Kubernetes- or Docker-Compose deployment file using the hub setup wizard of the Cryptomator website
4. for each vault already downloaded, re-download the vault template and overwrite the `vault.cryptomator` file at the destination with the newly downloaded `vault.cryptomator` of the template.

### DIFF
--- a/backend/src/main/java/org/cryptomator/hub/api/ConfigResource.java
+++ b/backend/src/main/java/org/cryptomator/hub/api/ConfigResource.java
@@ -26,7 +26,11 @@ public class ConfigResource {
 
 	@Inject
 	@ConfigProperty(name = "quarkus.oidc.client-id", defaultValue = "")
-	String keycloakClientId;
+	String keycloakClientIdHub;
+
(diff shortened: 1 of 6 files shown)


overheadhunter (approved):
Looks reasonable. I trust you have verified this change with the desktop app?

SailReal:
> I  trust you have verified this change with the desktop app?

Tested it ~300 times :joy:

PR #174, opened 2023-01-04
Use ENV variable to distinguish between self-hosted and managed instance

SailReal:
When deploying Hub, we can now set `HUB_MANAGED_INSTANCE=true` (default is `false`) to indicate if this is a self-hosted or a managed instance which in turn has an influence on the community license: managed instances does not have one.

Unfortunately I didn't get it running to have a test with a custom test profile inside a test with the default profile, that is the reason for the `BillingResourceManagedInstanceTest` class. Sad but wanted to test this case as well.

### DIFF
--- a/backend/src/main/java/org/cryptomator/hub/license/LicenseHolder.java
+++ b/backend/src/main/java/org/cryptomator/hub/license/LicenseHolder.java
@@ -75,19 +75,13 @@ public class LicenseHolder {
 	public boolean isExpired() {
 		return Optional.ofNullable(license) //
 				.map(l -> l.getExpiresAt().toInstant().isBefore(Instant.now())) //
-				.orElseGet(() -> {
-					if (!managedInstance) {
-						return CommunityLicenseConstants.IS_EXPIRED;
-					} else {
-						return ManagedInstanceNoLicenseConstants.IS_EXPIRED;
-					}
-				});
+				.orElse(false);
 	}
 
 	/**
 	 * Gets the number of available seats of the license
 	 *
-	 * @return Number of available seats, if license is not null. Otherwise {@value CommunityLicenseConstants#SEATS}.
+	 * @return Number of available seats, if license is not null. Otherwise {@value SelfHostedNoLicenseConstants#SEATS}.
 	 */
 	public long getAvailableSeats() {
 		return Optional.ofNullable(license) //
@@ -98,7 +92,7 @@ public class LicenseHolder {
 
 	public long getNoLicenseSeats() {
 		if (!managedInstance) {
-			return CommunityLicenseConstants.SEATS;
+			return SelfHostedNoLicenseConstants.SEATS;
 		} else {
 			return ManagedInstanceNoLicenseConstants.SEATS;
 		}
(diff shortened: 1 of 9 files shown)


backend/src/main/java/org/cryptomator/hub/license/LicenseHolder.java
```diff
@@ -69,7 +75,13 @@ public DecodedJWT get() {
...
 		return Optional.ofNullable(license) //
 				.map(l -> l.getExpiresAt().toInstant().isBefore(Instant.now())) //
-				.orElse(CommunityLicenseConstants.IS_EXPIRED);
+				.orElseGet(() -> {
+					if (!managedInstance) {
+						return CommunityLicenseConstants.IS_EXPIRED;
+					} else {
+						return ManagedInstanceNoLicenseConstants.IS_EXPIRED;
+					}
+				});
```
tobihagemann, line 84: I wonder if this is really necessary? If there is no license, it's always "not expired", right? Can't we just say `.orElse(false)`?

backend/src/main/java/org/cryptomator/hub/license/LicenseHolder.java
```diff
@@ -81,7 +93,19 @@ public long getAvailableSeats() {
...
+				.orElseGet(this::getNoLicenseSeats);
+	}
+
+	public long getNoLicenseSeats() {
+		if (!managedInstance) {
+			return CommunityLicenseConstants.SEATS;
+		} else {
+			return ManagedInstanceNoLicenseConstants.SEATS;
+		}
+	}
```
tobihagemann, line 105: How about renaming `CommunityLicenseConstants` to `SelfHostedNoLicenseConstants`? And also remove `Instance` in `ManagedInstanceNoLicenseConstants`?

frontend/src/i18n/en-US.json
```diff
@@ -37,9 +37,11 @@
   "admin.licenseInfo.expiresAt.description.valid": "Your license is valid.",
   "admin.licenseInfo.expiresAt.description.expired": "Your license has expired.",
   "admin.licenseInfo.manageSubscription": "Manage Subscription",
+  "admin.licenseInfo.type.title": "Type",
+  "admin.licenseInfo.upgradeLicense": "Get License",
```
tobihagemann, line 41: Rename `admin.licenseInfo.upgradeLicense` to `admin.licenseInfo.getLicense`?

frontend/src/i18n/en-US.json
```diff
@@ -37,9 +37,11 @@
...
   "admin.licenseInfo.communityLicense.description": "Thank you for using Cryptomator Hub! You have been granted the Community License. If you need more seats, upgrade your license.",
```
tobihagemann, line 42: Rename `admin.licenseInfo.communityLicense.description` to `admin.licenseInfo.selfHostedNoLicense.description`?

backend/src/main/java/org/cryptomator/hub/license/LicenseHolder.java
SailReal: Right, firstly Managed wasn't expired and after changing, I missed to change this block back, thx!

SailReal: Fixed in https://github.com/cryptomator/hub/pull/174/commits/9a7e1d46f452775cccec5351b8291a600f66f10e

frontend/src/i18n/en-US.json
SailReal: Fixed in https://github.com/cryptomator/hub/pull/174/commits/9a7e1d46f452775cccec5351b8291a600f66f10e

frontend/src/i18n/en-US.json
SailReal: Fixed in https://github.com/cryptomator/hub/pull/174/commits/9a7e1d46f452775cccec5351b8291a600f66f10e

PR #176, opened 2023-01-25
Recovery Key

overheadhunter:


frontend/src/components/CreateVault.vue
```diff
@@ -1,42 +1,82 @@
...
+              <p v-else class="text-sm text-red-900 mt-2">{{ t('createVault.error.invalidRecoveryKey') }}</p>
+            </div>
+          </div>
+        </div>
+      </div>
+    </form>
+  </div>
+
+  <div v-else-if="state == State.EnterVaultDetails">
+    <form ref="form" class="space-y-6" novalidate @submit.prevent="validateVaultDetails()">
```
infeo, line 42: Is it okay to bind different HTML forms to the same Vue ref?

frontend/src/i18n/de-DE.json
```diff
@@ -49,17 +52,27 @@
...
-  "createVault.vaultName": "Tresorname",
-  "createVault.vaultDescription": "Beschreibung",
-  "createVault.password": "Vault-Admin-Passwort",
-  "createVault.password.description": "Muss mindestens 8 Zeichen enthalten.",
-  "createVault.passwordConfirmation": "Vault-Admin-Passwort bestätigen",
-  "createVault.passwordConfirmation.passwordsMatch": "Passwörter stimmen überein!",
-  "createVault.passwordConfirmation.passwordsDoNotMatch": "Passwörter stimmen nicht überein",
-  "createVault.submit": "Tresor erstellen",
+  "createVault.enterRecoveryKey.title": "Tresor wiederherstellen",
+  "createVault.enterRecoveryKey.description": "Gib den Recovery-Key deines bestehenden Tresors ein, um daraus eine neue Tresorvorlage zu erstellen.",
```
infeo, line 56: In Cryptomator, the recovery key is called "Wiederherstellungsschlüssel" in the German translation. We should consider either renaming the hub translation or using "Recovery Key" as a specific, non-translatable term.

infeo (approved):
Looks shiny ✨ and works :D

frontend/src/components/CreateVault.vue
tobihagemann: In this case, it's not an issue because of the `v-if`s so that the Vue ref is always only used "once".

frontend/src/i18n/de-DE.json
tobihagemann: I'll give it a try again. The problem was that the word is just sooo looong. 😄

PR #181, opened 2023-02-17
Use more efficient primary key DDL

overheadhunter:
This migrates the primary key columns to [UUIDs](https://stackoverflow.com/a/32194327/4014509) where possible in order to [increase performance](https://stackoverflow.com/a/56453329/4014509), especially when joining.

User/Group IDs are managed externally and may not be valid UUIDs. Sames for device IDs, which are hex-encoded hashes of the device's public keys.

For now this only leaves us with the vault entity. Future additions should use UUIDs where possible.

---

~I also tried to set `COLLATE "C"` for `VARCHAR`-based ids, which should improve index performance, as chars are compared bitwise without any multi-byte character coding. However atm this is neither supported by Panache queries (see https://github.com/quarkusio/quarkus/issues/31254) nor by H2 (which wouldn't be a blocker though - just use postgres testcontainers for testing).~

Edit: No longer a blocker, since we ditched H2


overheadhunter:
some remarks for future additions: for test data I decided to start UUIDs with `7E57C0DE-0000-4000-8000-xxxxyyyyyyyy` where xxxx is the entity type (0001 being vaults) and yyyyyyyy can be used arbitrarily. For "invalid" or "not found" scenarios I set `xxxxyyyyyyyy` to `BADBADBADBAD`.

backend/src/test/java/org/cryptomator/hub/api/VaultResourceTest.java
```diff
@@ -596,20 +601,21 @@ public void getMembers3() {
 	@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
 	public class ManageGroups {
 
-		private final String vault4AdminJWT = "eyJhbGciOiJFUzM4NCIsInR5cCI6IkpXVCIsInZhdWx0SWQiOiJ2YXVsdDQifQ.eyJpYXQiOjE1MTYyMzkwMTUsImV4cCI6NTgxNjIzOTAzMCwibmJmIjoxNTE2MjM5MDAwfQ.aSe-P9j1vybUPS_Y1PDQ5knh_slaNulX365TSJNBYGY5dooUTZicu27_h_jYC-vvsafn4sjQhifEoXJq1U5r2kEp8ZV_Bn_4GdLLAG2JDewuAvWlkqks1un4spEO8yO4";
+											// eyJ2YXVsdElkIjoiN0U1N0MwREUtMDAwMC00MDAwLTgwMDAtMDAwMTAwMDA0NDQ0IiwidHlwIjoiSldUIiwiYWxnIjoiRVMzODQifQ.eyJpYXQiOjE2NzY2NTEwMTd9.HmbkbBO5yemWECLGwp8Q_dOEW1aoENkpZms5FFLhpPjTXb6cvCddBZF0ElK-6a4eG9Z_nR6RV0xsLA8BekBAjlTfYvqRwdQE5WJ2MzMtdxBYjs1Pic52jnJ0by8ko8Ap
```
infeo, line 604: What does this comment mean?

infeo (approved):
lgtm. Locally it runs fine. But i didn't tested the migration on an already running instance.

backend/src/test/java/org/cryptomator/hub/api/VaultResourceTest.java
overheadhunter: oh just some left-over code from testing. I'll remove it.

overheadhunter:
After merging #182, we are no longer restricted to ANSI SQL, allowing us to also set `COLLATE "C"` for `VARCHAR` fields knowing to contain just ASCII data, which [avoids respecting locale-specific rules for char comparison](https://stackoverflow.com/a/37549380/4014509) and instead does a simple binary comparison.

Note that the existing indices [don't need to be `REINDEX`ed](https://stackoverflow.com/a/38482103/4014509) after changing the table definition.

Interesting reads:
* https://gist.github.com/rraval/ef4e4bdc63e68fe3e83c9f98f56af7a4
* https://www.postgresql.eu/events/pgconfeu2022/sessions/session/4040/slides/337/Collations%20in%20PostgreSQL%20-%20The%20good,%20the%20bad%20and%20the%20ugly%20-%20Tobias%20Bussmann.pdf (slide 5.2 in particular)

backend/src/main/resources/org/cryptomator/hub/flyway/V4__Update_PKs.sql
```diff
@@ -0,0 +1,50 @@
...
+-- temporarily disable foreign key constraints
+ALTER TABLE "vault_access" DROP CONSTRAINT "VAULT_ACCESS_FK_VAULT";
+ALTER TABLE "access_token" DROP CONSTRAINT "ACCESS_FK_VAULT";
+ALTER TABLE "access_token" DROP CONSTRAINT "ACCESS_FK_DEVICE";
+
+-- collate C is more efficient on indexed columns (sadly user ids are generated externally and there is no guarantee for them to be UUIDs):
+ALTER TABLE "authority" ALTER COLUMN "id" SET DATA TYPE VARCHAR(255) COLLATE "C";
+ALTER TABLE "group_membership" ALTER COLUMN "group_id" SET DATA TYPE VARCHAR(255) COLLATE "C";
+ALTER TABLE "group_membership" ALTER COLUMN "member_id" SET DATA TYPE VARCHAR(255) COLLATE "C";
+ALTER TABLE "user_details" ALTER COLUMN "id" SET DATA TYPE VARCHAR(255) COLLATE "C";
```
infeo, line 16: If the user id is not generated by us, there is the possibility, that it contains non-ASCII chars.

No sane implementation would do this, but we should be aware of this (very small) risk.

backend/src/main/resources/org/cryptomator/hub/flyway/V4__Update_PKs.sql
```diff
@@ -0,0 +1,50 @@
...
+ALTER TABLE "group_details" ALTER COLUMN "id" SET DATA TYPE VARCHAR(255) COLLATE "C";
+ALTER TABLE "vault_access" ALTER COLUMN "authority_id" SET DATA TYPE VARCHAR(255) COLLATE "C";
+ALTER TABLE "device" ALTER COLUMN "owner_id" SET DATA TYPE VARCHAR(255) COLLATE "C";
+ALTER TABLE "device" ALTER COLUMN "id" SET DATA TYPE VARCHAR(64) COLLATE "C";
```
infeo, line 20: Why not use UUID for devices as well?

backend/src/main/resources/org/cryptomator/hub/flyway/V4__Update_PKs.sql
```diff
@@ -0,0 +1,50 @@
...
+ALTER TABLE "vault_access" ALTER COLUMN "vault_id" SET DATA TYPE UUID USING "vault_id"::uuid;
+ALTER TABLE "access_token" ALTER COLUMN "vault_id" SET DATA TYPE UUID USING "vault_id"::uuid;
+
+-- re-add constraints
+ALTER TABLE "vault_access" ADD CONSTRAINT "VAULT_ACCESS_FK_VAULT" FOREIGN KEY ("vault_id") REFERENCES "vault" ("id") ON DELETE CASCADE;
+ALTER TABLE "access_token" ADD CONSTRAINT "ACCESS_FK_DEVICE" FOREIGN KEY ("device_id") REFERENCES "device" ("id") ON DELETE CASCADE;
+ALTER TABLE "access_token" ADD CONSTRAINT "ACCESS_FK_VAULT" FOREIGN KEY ("vault_id") REFERENCES "vault" ("id") ON DELETE CASCADE;
+
+-- @formatter:off
+CREATE VIEW "effective_group_membership" ("group_id", "member_id", "path") AS
```
infeo, line 34: Just to ensure it exists afterwards, should we use `CREATE OR REPLACE` here?

backend/src/main/resources/org/cryptomator/hub/flyway/V4__Update_PKs.sql
```diff
@@ -0,0 +1,50 @@
...
+	SELECT "group_id", "member_id", 0, '/' || "group_id" || '/' || "member_id"
+	    FROM "group_membership"
+	UNION
+	SELECT "parent"."root", "child"."member_id", "parent"."depth" + 1, "parent"."path" || '/' || "child"."member_id"
+	    FROM "group_membership" "child"
+		INNER JOIN "members" "parent" ON "child"."group_id" = "parent"."member_id"
+		WHERE "parent"."depth" < 10
+) SELECT "root", "member_id", "path" FROM "members";
+
+CREATE VIEW "effective_vault_access" ("vault_id", "authority_id") AS
```
infeo, line 45: Just to ensure it exists afterwards, should we use CREATE OR REPLACE here?

infeo (changes requested):
Only to make it visible: We add `COLLATE` to colums, where it is not guaranteed that onyl ASCII characters are used. But since the column values are also external ids, we can safely assume that only ascii is used.

overheadhunter:
Even if multi-byte-chars are added to e.g. the user id column, it will still work. Collation only affects sorting.

backend/src/main/resources/org/cryptomator/hub/flyway/V4__Update_PKs.sql
overheadhunter: No. The view got dropped above. It must not exist. Let's be precise with assertions to not swallow unexpected states.

backend/src/main/resources/org/cryptomator/hub/flyway/V4__Update_PKs.sql
overheadhunter: See above.

backend/src/main/resources/org/cryptomator/hub/flyway/V4__Update_PKs.sql
overheadhunter: Existing entries are not UUIDs. Instead we have hex-encoded sha256 fingerprints of the device's public key.

infeo (approved):
No more open questions

PR #182, opened 2023-03-03
replace H2 with Postgres in `%test` profile

overheadhunter:
In order to use non-ANSI-SQL features, we need to make sure all profiles rely on the same dialect. So we need to drop h2 in unit tests, replacing it with Postgres (provided via testcontainers).

Sadly, this breaks `@FlywayTest`, so I needed to change a lot of unit tests to make sure, they don't modify anything prepared via `V9999__Test_Data.sql`. If a test wants to delete/change something, it should prepare its own data. As a general rule, a unit tests should "leave it (the database) as it found it". This is achieved in one of the following ways:

1. In some cases, `@TestTransaction` could be use to rollback such changes. This requires all to happen in the same transaction, obviously. So this is not feasible when using _REST assured_, as the serverside code is running outside the test transaction.
2. In other cases a `@AfterAll` function could do some cleanup.
3. For `@Order`ed tests creation and deletion could be run in pairs
4. If every other approach fails, the test itself needs to do the cleanup work. Ugly but simple.

Especially in `VaultResourceTest` I regrouped some of the nested tests and tried to make all tests run on the same set of common test data.

### DIFF
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -71,10 +71,6 @@
     </dependency>
     <dependency>
       <groupId>io.quarkus</groupId>
-      <artifactId>quarkus-jdbc-h2</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>io.quarkus</groupId>
       <artifactId>quarkus-jdbc-postgresql</artifactId>
     </dependency>
     <dependency>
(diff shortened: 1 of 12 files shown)


overheadhunter:
Btw: Diff is `+258 −383` 😎

SailReal (approved):
LGTM :+1:

I'm just thinking about where is the right place to document the options you have when writing tests, i.e. ultimately what you wrote in https://github.com/cryptomator/hub/pull/182#issue-1608941191, so that it doesn't get "lost" and you can directly take the best variant for the particular use case.

PR #185, opened 2023-03-09
`java.sql.Timestamp` → `java.time.Instant`

overheadhunter:
Make sure to always use the newer java.time API.

> You may exchange java.time objects directly with your database. Use a [JDBC driver](https://en.wikipedia.org/wiki/JDBC_driver) compliant with [JDBC 4.2](http://openjdk.java.net/jeps/170) or later. No need for strings, no need for java.sql.* classes.

(Source: [this StackOverflow answer](https://stackoverflow.com/a/51911760/4014509) by [someone who knows about timezones](https://stackoverflow.com/tags/timezone/topusers))

Also, make sure to store a `TIMESTAMP WITH TIME ZONE` in the database to avoid applying local date time to a timezoneless timestamp.

All persisted times should always be represented in UTC.


overheadhunter:
Btw: @infeo and I tested, how a database running in non-UTC timezone behaves, when applying the `ALTER TABLE` statement on existing data. Works as expected: Times before migration (represented in local time zone) are the same as after migration (still queried in local time zone unless expressed otherwise - but now stored in a way which allows querying from different time zones correctly).

infeo (approved):
looks good.

PR #192, opened 2023-04-25
Audit Logging

overheadhunter:
Proposed events to be added:

```ts
export type AuditEventDto = {
  id: number;
  timestamp: Date;
  type: 'CREATE_VAULT' | 'UNLOCK_VAULT' | 'UPDATE_VAULT_MEMBERSHIP';
}

export type CreateVaultEventDto = AuditEventDto & {
  userId: string;
  vaultId: string;
}

export type UnlockVaultEventDto = AuditEventDto & {
  userId: string;
  vaultId: string;
  deviceId: string;
  result: 'SUCCESS' | 'UNAUTHORIZED';
}

export type UpdateVaultMembershipEventDto = AuditEventDto & {
  userId: string;
  vaultId: string;
  authorityId: string;
  operation: 'ADD' | 'REMOVE';
}
```

Frontend should resolve IDs separately to avoid a single large query. This requires new backend APIs for resolving multiple IDs per resource type at once.


backend/src/main/java/org/cryptomator/hub/entities/UnlockResult.java
```diff
@@ -0,0 +1,7 @@
+package org.cryptomator.hub.entities;
+
+public enum UnlockResult {
+	SUCCESS,
+	NO_SUCH_DEVICE,
+	DEVICE_NOT_AUTHORIZED
+}
```
infeo, line 7: @overheadhunter Why is this enum standalone?

Also, similar to slf4j and its log levels, we can even add for each entry an nice, static convenience factory method:
```java
public static logSuccess(...) {
    log(..., SUCCESS);
}

public static logNoSuchDevice(...) {
    log(..., NO_SUCH_DEVICe);
}
...
```

backend/src/test/java/org/cryptomator/hub/api/AuditLogResourceTest.java
```diff
@@ -0,0 +1,31 @@
...
+	}
+
+	@Test
+	@TestSecurity(user = "Admin", roles = {"admin"})
+	@DisplayName("GET /auditlog?startDate=1970-01-01T00:00:03Z&endDate=1970-01-01T00:00:05Z returns 200 with two entries")
+	public void testGetAuditLogEntries() {
+		when().get("/auditlog?startDate=1970-01-01T00:00:03Z&endDate=1970-01-01T00:00:05Z")
+				.then().statusCode(200)
+				.body("id", Matchers.containsInAnyOrder(equalToIgnoringCase("7E57C0DE-0000-4000-8000-000510002222"), equalToIgnoringCase("7E57C0DE-0000-4000-8000-000510003333")));
+	}
```
overheadhunter, line 30: add another test with non-admin role, expecting access being denied

frontend/src/components/AuditLog.vue
```diff
@@ -0,0 +1,254 @@
...
+  return !isNaN(startDate.getTime());
+});
+const endDateFilterIsValid = computed(() => {
+  if (!/^\d{4}-\d{2}-\d{2}$/.test(endDateFilter.value)) {
+    return false;
+  }
+  const endDate = new Date(endDateFilter.value);
+  if (isNaN(endDate.getTime())) {
+    return false;
+  }
```
infeo, line 161: @tobihagemann can be dedup'ed

frontend/src/components/AuditLog.vue
```diff
@@ -0,0 +1,254 @@
...
+    const startDate = new Date(startDateFilter.value);
+    return startDate <= endDate;
+  }
+  return true;
+});
+const filterIsValid = computed(() => {
+  return startDateFilterIsValid.value && endDateFilterIsValid.value;
+});
+
+const currentPage = ref(1);
```
infeo, line 172: Better start with the logical numbering (zero based indexing) and just adjust the ui, otherwise you to always substract or add 1 in the code, decreasing readability.

frontend/src/components/AuditLog.vue
```diff
@@ -0,0 +1,254 @@
...
+
+onMounted(fetchData);
+
+async function fetchData() {
+  onFetchError.value = null;
+  try {
+    const events = await backend.auditLogs.getAllEvents(startDate.value, endDate.value, pages[currentPage.value - 1], pageSize.value + 1);
+    if (currentPage.value == 1 && pages[0] == 0 && events.length > 0) {
+      pages[0] = events[0].id - 1;
+    }
```
infeo, line 199: this is really hard to read/understand.

infeo (changes requested):
Some open questions. And the pagination part is... jeez, not so easy.

infeo (changes requested):
Another, but important thing

backend/src/main/resources/org/cryptomator/hub/flyway/V7__Audit_Events.sql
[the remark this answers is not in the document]
tobihagemann: That's what I thought as well (I created a seq in a previous commit) but did some reading on serial types:

https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-SERIAL

If we use `BIGSERIAL`, the seq is created implicitly.

backend/src/main/resources/org/cryptomator/hub/flyway/V1__Create_Tables.sql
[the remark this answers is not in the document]
tobihagemann: Fixes in f919a1baf2b092eee2ddad5a6c59d5c8da43da8b

frontend/src/components/AuditLog.vue
```diff
@@ -0,0 +1,254 @@
...
+    return 0;
+  }
+});
+const paginationEnd = computed(() => {
+  if (auditEvents.value) {
+    return (currentPage.value - 1) * pageSize.value + auditEvents.value.length;
+  } else {
+    return 0;
+  }
+});
```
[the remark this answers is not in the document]
tobihagemann: Good idea! But I would prefer it on a separate PR.

frontend/src/components/AuditLog.vue
tobihagemann: Fixed in d7fff0aa8bd4108d09dbfacb14315161dd90108f

frontend/src/components/AuditLog.vue
tobihagemann: Fixed in d7fff0aa8bd4108d09dbfacb14315161dd90108f

frontend/src/components/AuditLog.vue
tobihagemann: Fixed in e244d050b465efa1b1ffd8760b568843c20d7c9c

frontend/src/common/backend.ts
```diff
@@ -163,6 +163,52 @@ export type VersionDto = {
...
+  constructor(public id: number, public timestamp: Date, public type: AuditEventType) { }
+
+  abstract details(): { [key: string]: string };
+}
+
+enum UnlockResult {
+  Success = 'SUCCESS',
+  NoSuchDevice = 'NO_SUCH_DEVICE',
+  DeviceNotAuthorized = 'DEVICE_NOT_AUTHORIZED'
+}
```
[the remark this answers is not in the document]
tobihagemann: Interesting... it looks like that TypeScript enums are transpiled into specific JavaScript code, which increases the generated JavaScript code in size, which doesn't happen with string literals. The only drawback is the lack of support for auto-completion in the IDE, I guess.

Maybe we can even scrap all of that completely and just handle the unlock result as a string since it is not really "used" at the moment. 😄

backend/src/main/resources/org/cryptomator/hub/flyway/V7__Audit_Events.sql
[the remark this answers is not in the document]
overheadhunter: Oh I see, so the sequence name is just a convention? Or can we set it explicitly?

backend/src/main/java/org/cryptomator/hub/entities/UnlockResult.java
infeo: I read your comment and fixed it in https://github.com/cryptomator/hub/pull/192/commits/4532f8badd8bb3f5459cdb67dca7b83a6f46defd

backend/src/main/java/org/cryptomator/hub/entities/UnlockResult.java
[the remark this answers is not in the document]
infeo: Changed in https://github.com/cryptomator/hub/pull/192/commits/4532f8badd8bb3f5459cdb67dca7b83a6f46defd

backend/src/main/java/org/cryptomator/hub/api/AuditLogResource.java
```diff
@@ -0,0 +1,71 @@
...
+@Path("/auditlog")
+public class AuditLogResource {
+
+	@GET
+	@RolesAllowed("admin")
+	@Produces(MediaType.APPLICATION_JSON)
+	@Operation(summary = "list all auditlog entries within a period", description = "list all auditlog entries from a period specified by a start and end date")
+	@Parameter(name = "startDate", description = "the start date of the period as ISO 8601 datetime string", in = ParameterIn.QUERY)
+	@Parameter(name = "endDate", description = "the end date of the period as ISO 8601 datetime string", in = ParameterIn.QUERY)
+	@Parameter(name = "after", description = "the id of the last entry of the previous page", in = ParameterIn.QUERY)
```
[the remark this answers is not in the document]
infeo: I renamed it in https://github.com/cryptomator/hub/pull/192/commits/7eec043631a44b728080414c5d13f7916c50859c to `afterId` and updated the doc to not what we are using it for, but what it actually does.

backend/src/test/java/org/cryptomator/hub/api/AuditLogResourceTest.java
infeo: Fixed in https://github.com/cryptomator/hub/pull/192/commits/89e515472e28c4c7917d501f0c42496c4f3c2ff1

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
```diff
@@ -239,10 +241,13 @@ public String unlock(@PathParam("vaultId") UUID vaultId, @PathParam("deviceId")
 
 		var access = AccessToken.unlock(vaultId, deviceId, jwt.getSubject());
 		if (access != null) {
+			UnlockEvent.log(jwt.getSubject(), vaultId, deviceId, UnlockResult.SUCCESS);
 			return access.jwe;
 		} else if (Device.findById(deviceId) == null) {
+			UnlockEvent.log(jwt.getSubject(), vaultId, deviceId, UnlockResult.NO_SUCH_DEVICE);
 			throw new NotFoundException("No such device.");
 		} else {
+			UnlockEvent.log(jwt.getSubject(), vaultId, deviceId, UnlockResult.DEVICE_NOT_AUTHORIZED);
```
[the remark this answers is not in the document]
infeo: If we remove it now already, but decide to make another release before the unlock-change is merged, users cannot register their devices anymore. I would keep it and make the change during or after the unlock-refactoring.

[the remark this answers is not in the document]
overheadhunter: Removing a log event does not prevent people from registering devices

frontend/src/common/backend.ts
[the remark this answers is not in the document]
overheadhunter: > The only drawback is the lack of support for auto-completion in the IDE, I guess.

VSCode has no problem autocompleting such strings, btw.

Bildschirmfoto 2023-05-15 um 12 30 14

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
```diff
@@ -8,28 +8,10 @@
...
-import jakarta.ws.rs.DELETE;
-import jakarta.ws.rs.ForbiddenException;
-import jakarta.ws.rs.GET;
-import jakarta.ws.rs.InternalServerErrorException;
-import jakarta.ws.rs.NotFoundException;
-import jakarta.ws.rs.PUT;
-import jakarta.ws.rs.Path;
-import jakarta.ws.rs.PathParam;
-import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.*;
```
overheadhunter, line 11: Unnecessary diff. Don't we have code style settings shared in this repo?

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
```diff
@@ -8,28 +8,10 @@
...
 import jakarta.ws.rs.core.MediaType;
 import jakarta.ws.rs.core.Response;
-import org.cryptomator.hub.entities.AccessToken;
-import org.cryptomator.hub.entities.Device;
-import org.cryptomator.hub.entities.EffectiveGroupMembership;
-import org.cryptomator.hub.entities.EffectiveVaultAccess;
-import org.cryptomator.hub.entities.Group;
-import org.cryptomator.hub.entities.User;
-import org.cryptomator.hub.entities.Vault;
+import org.cryptomator.hub.entities.*;
```
overheadhunter, line 14: see above

backend/src/main/java/org/cryptomator/hub/entities/UnlockEvent.java
```diff
@@ -0,0 +1,67 @@
...
+		event.userId = userId;
+		event.vaultId = vaultId;
+		event.deviceId = deviceId;
+		event.result = result;
+		event.persist();
+	}
+
+	public enum Result {
+		SUCCESS,
+		NO_SUCH_DEVICE,
```
overheadhunter, line 63: to be removed

backend/src/test/resources/org/cryptomator/hub/flyway/V9999__Test_Data.sql
```diff
@@ -55,3 +55,17 @@ VALUES
 	('device1', '7E57C0DE-0000-4000-8000-000100001111', 'jwe1'),
 	('device2', '7E57C0DE-0000-4000-8000-000100001111', 'jwe2'),
 	('device3', '7E57C0DE-0000-4000-8000-000100002222', 'jwe3'); -- device3 of user1, part of group1, has access to vault2
+
+INSERT INTO "audit_event" ("id", "timestamp", "type")
+VALUES
+    (999, '1970-01-01T00:00:02.900Z', 'UNLOCK'),
+    (1000, '1970-01-01T00:00:03Z', 'UNLOCK'),
+    (1111, '1970-01-01T00:00:04Z', 'UNLOCK'),
+    (4242, '1970-01-01T00:00:05Z', 'UNLOCK');
```
overheadhunter, line 64: since we try to use some "speaking" IDs in the test data, let's try to find some better values.

frontend/src/router/index.ts
```diff
@@ -79,6 +80,15 @@ const routes: RouteRecordRaw[] = [
...
         }
       },
+      {
+        path: 'auditlog',
+        component: AuditLog,
+        props: (route) => ({ token: route.query.token }),
+        beforeEnter: async () => {
+          const auth = await authPromise;
+          return auth.isAdmin(); //TODO: reroute to NotFound Screen/ AccessDeniedScreen?
+        }
```
overheadhunter, line 90: instead of repeating this on every admin resource, shouldn't we rather add some nesting? Using paths such as `/admin/auditlog` and `/admin/settings`, with the latter being the default when visiting `/admin/`.

backend/src/main/resources/org/cryptomator/hub/flyway/V7__Audit_Events.sql
```diff
@@ -0,0 +1,20 @@
+-- Cryptomator Hub 1.1.0
+
+CREATE TABLE "audit_event"
+(
+	"id"        BIGSERIAL,
```
overheadhunter, line 5: I would prefer the "long" sequence declaration (see mentioned link in https://github.com/cryptomator/hub/pull/192#discussion_r1190883179), so we can _explicitly_ choose a sequence name in a `CREATE SEQUENCE` statement, which will then be used in `AuditEvent.java`'s `@SequenceGenerator`.

backend/src/test/resources/org/cryptomator/hub/flyway/V9999__Test_Data.sql
infeo: @overheadhunter Do you have any suggestios? Due to the datatype we have to use a number. We can use hex notation to "write" something, but apart from that, i can't imagine to make the ids more "speakable".

The reason for picking such high numbers is, that other integrations-tests might fail using "unlock" might fail, because our sequence generator begins at 1 (or zero), but the table contains already entries for those low ids.

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
infeo: Fixed in https://github.com/cryptomator/hub/pull/192/commits/7e57dcf691673b5303890423dfe3155ede18febf

backend/src/main/java/org/cryptomator/hub/entities/UnlockEvent.java
infeo: Fixed in https://github.com/cryptomator/hub/pull/192/commits/7e57dcf691673b5303890423dfe3155ede18febf

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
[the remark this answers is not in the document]
infeo: Fixed in https://github.com/cryptomator/hub/pull/192/commits/7e57dcf691673b5303890423dfe3155ede18febf

frontend/src/common/backend.ts
```diff
@@ -163,6 +163,52 @@ export type VersionDto = {
...
+      && typeof unlockEventDto.vaultId === 'string'
+      && typeof unlockEventDto.deviceId === 'string'
+      && Object.values(UnlockResult).includes(unlockEventDto.result)
+      && unlockEventDto.type === AuditEventType.Unlock;
+  }
+
+  static copy(obj: UnlockEventDto): UnlockEventDto {
+    return new UnlockEventDto(obj.id, new Date(obj.timestamp), obj.type, obj.userId, obj.vaultId, obj.deviceId, obj.result);
+  }
+}
```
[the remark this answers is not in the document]
tobihagemann: Your suggestion didn't quite work because `details` is not part of the DTO. It was a method I added that only existed in the frontend code. In the backend, `UnlockEventDto` is an extension of `AuditEventDto`.

My goal was to show only the key/values of the subtype in the details column.

I did the refactoring in affd91798e270961e87804fb03a3e88cdc4f5613. Does it make sense?

- `AuditEventDto` is now a simple type and `UnlockEventDto` is an extension of `AuditEventDto` (by extending the type via intersections).
- `AuditLogService.getAllEvents()` doesn't map anything specific anymore.
- `AuditLog` component now has some code to compute the `auditEventDetails`, which transforms the audit events by filtering out the `AuditEventDto` properties.

An alternative would be to somehow change the backend to return a DTO that matches your suggestion. But I didn't know if that would make sense and if it would complicate things in the backend. I'm open for suggestions.

backend/src/main/resources/org/cryptomator/hub/flyway/V7__Audit_Events.sql
tobihagemann: Fixed in 7f14e11217ec6396448ec46008f6a3ff9af6fd2d

frontend/src/common/backend.ts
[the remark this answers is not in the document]
tobihagemann: Fixed in affd91798e270961e87804fb03a3e88cdc4f5613

backend/src/test/resources/org/cryptomator/hub/flyway/V9999__Test_Data.sql
overheadhunter: Should not keep us from merging. Test data can be adjusted at any time, since this is no "real" migration.

frontend/src/common/backend.ts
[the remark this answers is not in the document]
overheadhunter: current impl looks very clean

frontend/src/common/backend.ts
```diff
@@ -166,7 +167,90 @@ export type VersionDto = {
...
+export type UpdateVaultMembershipEventDto = AuditEventDto & {
+  userId: string;
+  vaultId: string;
+  authorityId: string;
+  operation: 'ADD' | 'REMOVE';
+}
+
+/* AuditLogEntityCache */
+
+export class AuditLogEntityCache {
```
overheadhunter, line 194: this should get an `invalidate` method, that is called upon click on "refresh" in the audit log UI

backend/src/test/resources/org/cryptomator/hub/flyway/V9999__Test_Data.sql
```diff
@@ -55,3 +55,17 @@ VALUES
...
+
+INSERT INTO "unlock_event" ("id", "user_id", "vault_id", "device_id", "result")
```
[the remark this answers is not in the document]
overheadhunter: oops this is an old comment, I don't know why it hasn't been posted some days ago...

frontend/src/common/backend.ts
tobihagemann: Fixed in 6ab57deca6f103259b5f2cfde6738c8957d52159

infeo (approved):

PR #193, opened 2023-04-25
changed column type for "iteration count" from varchar(255) to integer

overheadhunter:
note that this Flyway migration is V7. I assume this will get merged after #192 (V6)


SailReal (approved):
LGTM :+1:

PR #194, opened 2023-04-27
Update dependencies

SailReal:


backend/src/main/java/org/cryptomator/hub/api/DeviceResource.java
```diff
@@ -57,7 +58,7 @@ public Response create(@Valid DeviceDto deviceDto, @PathParam("deviceId") @Valid
 			device.persistAndFlush();
 			return Response.created(URI.create(".")).build();
 		} catch (PersistenceException e) {
-			if (e.getCause() instanceof ConstraintViolationException) {
+			if (e instanceof ConstraintViolationException) {
```
overheadhunter, line 61: just for the record, because this got me wondering: With Hibernate 6, [`HibernateException` is now finally a `PersistenceException`](https://github.com/hibernate/hibernate-orm/commit/8ef423a1b3bd315724358513162a94bbba4f8822).

overheadhunter (approved):

Issue #199, opened 2023-05-22
Incomplete openapi for devices-requiring-access-grant due to missing 200 response code declaration


chenkins:
### Please agree to the following

- [X] I have searched [existing issues](https://github.com/cryptomator/hub/issues?q=) for duplicates
- [X] I agree to follow this project's [Code of Conduct](https://github.com/cryptomator/hub/blob/develop/.github/CODE_OF_CONDUCT.md)

### Summary

Incomplete OpenApi documentation 

### System Setup

```markdown
- Hub: 1.2.0-SNAPSHOT
- Keycloak: 21.1.1
```

### Steps to Reproduce

1. Get the json from `http://localhost:8080/q/openapi?format=json` 

### Expected Behavior

```
"/api/vaults/{vaultId}/devices-requiring-access-grant" : {
      "get" : {
        "tags" : [ "Vault Resource" ],
        "summary" : "list devices requiring access rights",
        "description" : "lists all devices owned by vault members, that don't have a device-specific masterkey yet",
        "parameters" : [ {
          "name" : "vaultId",
          "in" : "path",
          "required" : true,
          "schema" : {
            "$ref" : "#/components/schemas/UUID"
          }
        } ],
        "responses" : {
          "200" : {
            "description" : "OK",
            "content" : {
              "application/json" : {
                "schema" : {
                  "type" : "array",
                  "items" : {
                    "$ref" : "#/components/schemas/DeviceDto"
                  }
                }
              }
            }
          },
          "401" : {
            "description" : "VaultAdminAuthorizationJWT not provided"
          },
          "403" : {
            "description" : "VaultAdminAuthorizationJWT expired or not yet valid"
          },
          "404" : {
            "description" : "vault not found"
          }
        },
        "security" : [ {
          "SecurityScheme" : [ "user" ]
        } ]
      }
    },
```    

### Actual Behavior

```
"/api/vaults/{vaultId}/devices-requiring-access-grant" : {
      "get" : {
        "tags" : [ "Vault Resource" ],
        "summary" : "list devices requiring access rights",
        "description" : "lists all devices owned by vault members, that don't have a device-specific masterkey yet",
        "parameters" : [ {
          "name" : "vaultId",
          "in" : "path",
          "required" : true,
          "schema" : {
            "$ref" : "#/components/schemas/UUID"
          }
        } ],
        "responses" : {
          "401" : {
            "description" : "VaultAdminAuthorizationJWT not provided"
          },
          "403" : {
[...]

PR #200, opened 2023-05-22
#199 declare missing 200 APIResponse

chenkins:


infeo:
@chenkins Thanks for your contribution!

PR #202, opened 2023-06-07
Feature: Archive vault

### LINKED ISSUE #112: Archive vaults (Not delete)
Vaults cannot be deleted to not loose any user data. Instead, vaults should be archived: The vault does not appear in the normal vault list, cannot be regularly managed and it cannot be used for unlocking.

Is a vault archived, one should be able to activate it again.

infeo:
Closes #112.

This PR adds the funtionality to archive a vault. ~~Additionally, the REST endpoint for achieving it let the user also update the vault name and description.~~ The vault is not deleted, it is only marked as archived in the database. The list of members is kept.

To implement functionality the the `/vaults/{vaultid}` was refactored. It now accepts PUT requests for already existing vaults, but then _selectively_ updates it. Updatable are name, description and archive state.

For archived vaults in the backend the following rules hold:
* they are not listed in `getAccessible()`
* they are listed in `getAllVaults()` and `getSomeVaults()`
* requests to `unlock()` are blocked and return 403
* all other requests are still possible

As noted by @stefan-karg, users in archived vaults do not increase the license seats. Since the member lists are retained, reactivating a vault can impact the license up to the point that any unlock-operation is blocked because the license limit is exceeded.

A database migration is required which updates the vault table by a new column and filters archived vaults from the effective_vault_access table.

TODOS:
- [x] Review backend
- [x] implement frontend


infeo:
@overheadhunter I simplfied the backend according to your suggestions.

infeo:
Screenshots of the frontend:
* Vault list
grafik

* vault details
grafik

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
```diff
@@ -348,17 +358,50 @@ public Response create(@PathParam("vaultId") UUID vaultId, @Valid VaultDto vault
...
+		if (update.description != null) {
+			vault.description = update.description;
+		}
+
+		if (update.archived != null) {
+			if (!("false".equalsIgnoreCase(update.archived) || "true".equalsIgnoreCase(update.archived))) {
+				throw new BadRequestException("Allowed values for field \"archived\" are [true, false]");
+			}
+			vault.archived = Boolean.valueOf(update.archived);
+		}
```
overheadhunter, line 386: I dislike the `if x != null` part.

The request body shall represent the whole truth. In ReST APIs I don't expect the backend to reinterpret it. Updating must include removing one or more of the resource's properties (maybe I want to delete a vault's description).

If the frontend likes to keep things unchanged, it should send them unchanged, not exclude them from a DTO.

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
```diff
@@ -372,8 +415,12 @@ public Vault toVault(UUID id) {
 			vault.salt = salt;
 			vault.authenticationPublicKey = authPublicKey;
 			vault.authenticationPrivateKey = authPrivateKey;
+			vault.archived = archived;
 			return vault;
 		}
+	}
+
+	public record VaultUpdateDto(@JsonProperty("name") String name, @JsonProperty("description") String description, @JsonProperty("archived") String archived) {
```
overheadhunter, line 423: Don't we just want to send an updated VaultDTO?

backend/src/test/java/org/cryptomator/hub/api/VaultResourceTest.java
```diff
@@ -585,7 +590,8 @@ public void setup() throws SQLException {
 							('user91'),
 							('user92'),
 							('user93'),
-							('user94');
+							('user94'),
+							('userAA');
```
overheadhunter, line 594: Do we really need a different user?

If yes: do we really need a different naming scheme for the IDs?

frontend/src/common/backend.ts
```diff
@@ -49,6 +50,12 @@ export type VaultDto = {
   authPrivateKey: string;
 };
 
+export type VaultUpdateDto = {
+  name?: string;
+  description?: string;
+  archived?: string 
+};
+
```
overheadhunter, line 58: See above: can't we reuse the VaultDto?

frontend/src/components/VaultDetails.vue
```diff
@@ -8,12 +8,20 @@
...
+        <h4 class="ml-3 text-sm font-medium text-yellow-800">This vault is archived</h4>
+      </div>
+    </div>
     <div>
       <h3 class="font-medium text-gray-900">{{ t('vaultDetails.description.header') }}</h3>
       <div class="mt-2 flex items-center justify-between">
         <p v-if="vault != null && vault.description.length > 0" class="text-sm text-gray-500">{{ vault.description }}</p>
         <p v-else class="text-sm text-gray-500 italic">{{ t('vaultDetails.description.empty') }}</p>
-        <!-- TODO: add rest API to change vault metadata in backend
+        <!-- TODO: implement calls to change metadata (see backend.ts -> VaultService - update())
```
overheadhunter, line 24: Done?

overheadhunter (changes requested):

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
```diff
@@ -348,17 +358,50 @@ public Response create(@PathParam("vaultId") UUID vaultId, @Valid VaultDto vault
 		}
 	}
 
+	@PATCH
```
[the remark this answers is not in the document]
infeo: I deliberately used PATCH, because `PUT` is used for creating a vault:
https://github.com/cryptomator/hub/blob/7aa07c12fad67fdfea04af89639a57d72680d721/backend/src/main/java/org/cryptomator/hub/api/VaultResource.java#L328-L330

According to the specs we cannot put the same vault twice. I guess the reason was that PUT replaces an object, meaning we could also replace (in the code) a masterkey of a vault. Of course, like you mentioned earlier, we can make the backend dumb and just accept what we put. But then we would have the non-zero risk of destroying keymaterial.

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
infeo: That is part of the PATCH contract, at least to the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH):
>A PATCH request is considered a set of instructions on how to modify a resource. Contrast this with [PUT](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT); which is a complete representation of a resource.

backend/src/test/java/org/cryptomator/hub/api/VaultResourceTest.java
infeo: Yes, because this user is part of an archived vault. I deliberately changed the naming scheme to indcate what the user is used for.

infeo: But i guess i can adapt more the naming scheme. Will change it.

frontend/src/components/VaultDetails.vue
infeo: Nope, since the actual calls in `VaultDetails.vue` are not implemented. I wanted to seperate these matters.

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
overheadhunter: I don't think that by "complete" the spec means the same as non-null in Java. Imagine a resource with the initial state `{"a": 1, "b": null, "c": "hello"}`. Now let's patch it using `{"b": "world", "c:" null}`. I would expect the resource to be `{"a": 1, "b": "world", "c": null}` now.

However, the current approach doesn't allow us to distinguish between "I want a field to be null" and "I don't want to modify a field", partly due to how Java/Jackson work.

If we instead contractually specify that we always send the "full" json representation in the request body, we don't have to deal with "missing" entries. If the sender doesn't want a value to change, he must not change it (but still send it). Furthermore, this allows better protection against race conditions: If the sender sends `{"a": 1, ...}` but the resource has been changed to `a=2` on the server, we can easily implement server-side checks and respond with 428. If however, `a` was omitted, the server would update the resource, leaving the client under the false impression that `a` would still be `1`.

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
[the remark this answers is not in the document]
overheadhunter: The server is still free to reject any changes to existing objects that it doesn't like. Or is there any spec that says that a PUT _must_ overwrite existing data, no questions asked?

backend/src/test/java/org/cryptomator/hub/api/VaultResourceTest.java
overheadhunter: I would really like to keep it consistent, keeping different number ranges for different test cases. With the "low numbers", defined in `V9999__Test_Data.sql` being "reusable", so they should only be access for read-only test cases. Higher numbers being test-specific, volatile and should not overlap.

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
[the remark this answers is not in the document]
infeo: > The server is still free to reject any changes to existing objects that it doesn't like. Or is there any spec that says that a PUT must overwrite existing data, no questions asked?

Of course not, but that would add logic to our backend again, instead of having a dumb one. And how should the server react on fields that should not change? Should those be ignored or a bad request code returned? Logged at least? And we also would drop the 409 code, because there is no distinction possible anymore between "update" and "already exists".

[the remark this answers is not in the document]
infeo: After a discussion we have a plan now: The  `/vaults/{vaultid}` PATCH endpoint will be removed again. The update functionality will be merged into the respective PUT endpoint. Done in https://github.com/cryptomator/hub/pull/202/commits/a21369f2be3497e5d8456a3ff9a1761ae1d85358

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
infeo: Obsolete due to https://github.com/cryptomator/hub/pull/202#discussion_r1231961355

frontend/src/common/backend.ts
infeo: removed in https://github.com/cryptomator/hub/pull/202/commits/861fc8b08fe23249fc2c8501c7b2c3a711365127

infeo:
@tobihagemann I saw that https://github.com/cryptomator/hub/pull/203/files also requires a db migration. Should i rename the migration of this PR directly to V9 ?

tobihagemann:
@infeo Yes, that would make sense.

frontend/src/components/ArchiveVaultDialog.vue
```diff
@@ -0,0 +1,99 @@
...
+  const v = props.vault;
+  try {
+    const updatedVault = await backend.vaults.createOrUpdateVault(v.id, v.name, v.description, true, v.masterkey, v.iterations, v.salt, v.authPublicKey, v.authPrivateKey );
+    if (updatedVault != null){
+      emit('archived', updatedVault);
+    } else {
+      //TODO: or throw error in this case? But the "damage" is already done in backend
+      const createdVault: VaultDto = { id: v.id, name: v.name, description: v.description, archived: true, creationTime: v.creationTime, masterkey: v.masterkey, iterations: v.iterations, salt: v.salt, authPublicKey: v.authPublicKey, authPrivateKey: v.authPrivateKey };
+      emit('archived', createdVault);
+    }
```
tobihagemann, line 91: I'm wondering about these lines of code (and also the same ones in `ReactivateVaultDialog`). Do we really want to make the distinction between "create" and "update" by a missing/existing response object? If yes, this should be documented in the backend API.

And, honestly, I think your TODO includes a legitimate question. What do we even do if a new vault has been created instead? In theory, we can ignore this case because vaults can't be deleted at the moment. But that doesn't have to be the case forever.

tobihagemann (changes requested):

frontend/src/components/ArchiveVaultDialog.vue
infeo: >Do we really want to make the distinction between "create" and "update" by a missing/existing response object? 

Good question! I'll change the code to always return the object, will also clean up code. Creation/Update can still be differentiated with the response code.

> What do we even do if a new vault has been created instead? In theory, we can ignore this case because vaults can't be deleted at the moment. But that doesn't have to be the case forever.

Had the same thoughts. In my opinion, even for the long term this is okay, because deletion should really be the last option. If we ever allow deletion, we have to revise more than just archiving/reactivating anyway.

tobihagemann (approved):

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
```diff
@@ -247,10 +248,15 @@ public List<DeviceResource.DeviceDto> getDevicesRequiringAccessGrant(@PathParam(
 	@Operation(summary = "get the device-specific masterkey")
 	@APIResponse(responseCode = "200")
 	@APIResponse(responseCode = "402", description = "number of effective vault users exceeds available license seats")
-	@APIResponse(responseCode = "403", description = "device not authorized to access this vault")
-	@APIResponse(responseCode = "404", description = "unknown device")
+	@APIResponse(responseCode = "403", description = "vault is archived or device not authorized to access this vault")
+	@APIResponse(responseCode = "404", description = "vault or device not found")
```
overheadhunter, line 252: while physically still present, I would argue that after archiving, the vault does no longer "exist" from an "unlock perspective", hence I would rather throw a NotFoundException.

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
```diff
@@ -325,22 +331,43 @@ public VaultDto get(@PathParam("vaultId") UUID vaultId) {
...
-	@Operation(summary = "creates a vault",
-			description = "Creates a vault with the given vault id. The creationTime in the vaultDto is ignored and the current server time is used.")
-	@APIResponse(responseCode = "201", description = "vault created")
-	@APIResponse(responseCode = "409", description = "vault with given id or name already exists")
-	public Response create(@PathParam("vaultId") UUID vaultId, @Valid VaultDto vaultDto) {
-		if (vaultDto == null) {
-			throw new BadRequestException("Missing vault dto");
+	@Operation(summary = "creates or updates a vault",
+			description = "Creates or updates a vault with the given vault id. The creationTime in the vaultDto is always ignored. On creation the current server time is used. On update, only the name, description and archived fields are considered.")
+	@APIResponse(responseCode = "200", description = "existing vault updated")
```
overheadhunter, line 336: 200 or 204 are both fine, I don't have any preference. However I would like all our resources to behave the same way. Have you taken a look at other "update-PUTs"?

see https://stackoverflow.com/a/2342589/4014509

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
```diff
@@ -325,22 +331,43 @@ public VaultDto get(@PathParam("vaultId") UUID vaultId) {
...
+		vault.archived = vaultDto.archived;
+
 		try {
 			vault.persistAndFlush();
 			CreateVaultEvent.log(currentUser.id, vault.id);
-			return Response.created(URI.create(".")).build();
+			if (isCreated) {
+				return Response.created(URI.create(".")).contentLocation(URI.create(".")).entity(VaultDto.fromEntity(vault)).type(MediaType.APPLICATION_JSON).build();
+			} else {
+				return Response.ok(VaultDto.fromEntity(vault), MediaType.APPLICATION_JSON).build();
```
overheadhunter, line 369: Ok, in case of a response body, status code 200 is probably the correct choice.

backend/src/test/java/org/cryptomator/hub/api/VaultResourceTest.java
```diff
@@ -69,9 +71,11 @@ public static void beforeAll() throws NoSuchAlgorithmException, InvalidKeySpecEx
...
 		var algorithmVault2 = Algorithm.ECDSA384((ECPrivateKey) getPrivateKey("MIG2AgEAMBAGByqGSM49AgEGBSuBBAAiBIGeMIGbAgEBBDCAHpFQ62QnGCEvYh/pE9QmR1C9aLcDItRbslbmhen/h1tt8AyMhskeenT+rAyyPhGhZANiAAQLW5ZJePZzMIPAxMtZXkEWbDF0zo9f2n4+T1h/2sh/fviblc/VTyrv10GEtIi5qiOy85Pf1RRw8lE5IPUWpgu553SteKigiKLUPeNpbqmYZUkWGh3MLfVzLmx85ii2vMU="));
+		var algorithmVaultArchived = Algorithm.ECDSA384((ECPrivateKey) getPrivateKey("MIG2AgEAMBAGByqGSM49AgEGBSuBBAAiBIGeMIGbAgEBBDCAHpFQ62QnGCEvYh/pE9QmR1C9aLcDItRbslbmhen/h1tt8AyMhskeenT+rAyyPhGhZANiAAQLW5ZJePZzMIPAxMtZXkEWbDF0zo9f2n4+T1h/2sh/fviblc/VTyrv10GEtIi5qiOy85Pf1RRw8lE5IPUWpgu553SteKigiKLUPeNpbqmYZUkWGh3MLfVzLmx85ii2vMU="));
 
 		vault1AdminJWT = JWT.create().withHeader(Map.of("vaultId", "7E57C0DE-0000-4000-8000-000100001111")).withIssuedAt(Instant.now()).sign(algorithmVault1);
 		vault2AdminJWT = JWT.create().withHeader(Map.of("vaultId", "7E57C0DE-0000-4000-8000-000100002222")).withIssuedAt(Instant.now()).sign(algorithmVault2);
+		vaultArchivedAdminJWT = JWT.create().withHeader(Map.of("vaultId", "7E57C0DE-0000-4000-8000-AAAAAAAAAAAA")).withIssuedAt(Instant.now()).sign(algorithmVaultArchived);
```
overheadhunter, line 78: `7E57C0DE-0000-4000-8000-0001` is the UUID prefix we use for "vaults". Not documented, my fault. But this way we can use different namespaces for test data.

```suggestion
		vaultArchivedAdminJWT = JWT.create().withHeader(Map.of("vaultId", "7E57C0DE-0000-4000-8000-0001AAAAAAAA")).withIssuedAt(Instant.now()).sign(algorithmVaultArchived);
```

backend/src/test/java/org/cryptomator/hub/api/VaultResourceTest.java
```diff
@@ -683,6 +652,62 @@ public void testGet(String method, String path) {
...
+	public class Unlock {
+
+		@Test
+		@DisplayName("GET /vaults/{vaultid}/keys/{deviceId} returns 404 for not-existing vaults")
+		@TestSecurity(user = "User Name 1", roles = {"user", "admin"})
+		@OidcSecurity(claims = {
+				@Claim(key = "sub", value = "user1")
+		})
+		public void testUnlockNotExistingVault() {
+			when().get("/vaults/{vaultId}/keys/{deviceId}", "7E57C0DE-0000-4000-8000-FFFFFFFFFFFF", "someDevice")
```
overheadhunter, line 694: use BAD ids for non-existing or otherwise broken/invalid test data

```suggestion
			when().get("/vaults/{vaultId}/keys/{deviceId}", "7E57C0DE-0000-4000-8000-BADBADBADBAD", "someDevice")
```

backend/src/test/java/org/cryptomator/hub/api/VaultResourceTest.java
```diff
@@ -723,17 +748,97 @@ public void testListSomeVaultsNoParams() {
...
+
+		@AfterAll
+		public void deleteData() throws SQLException {
+			try (var s = dataSource.getConnection().createStatement()) {
+				s.execute("""
+						DELETE FROM "vault"
+						WHERE "id" IN ('7E57C0DE-0000-4000-8000-FFFFFFFF1111','7E57C0DE-0000-4000-8000-FFFFFFFF2222');
+						""");
+			}
+		}
```
overheadhunter, line 841: 0001 id namespace

frontend/src/common/backend.ts
```diff
@@ -323,9 +324,10 @@ class VaultService {
       .then(response => response.data).catch(err => rethrowAndConvertIfExpected(err, 403));
   }
 
-  public async createVault(vaultId: string, name: string, description: string, masterkey: string, iterations: number, salt: string, signPubKey: string, signPrvKey: string): Promise<AxiosResponse<any>> {
-    const body: VaultDto = { id: vaultId, name: name, description: description, creationTime: new Date(), masterkey: masterkey, iterations: iterations, salt: salt, authPublicKey: signPubKey, authPrivateKey: signPrvKey };
+  public async createOrUpdateVault(vaultId: string, name: string, description: string, archived: boolean, masterkey: string, iterations: number, salt: string, signPubKey: string, signPrvKey: string): Promise<VaultDto> {
+    const body: VaultDto = { id: vaultId, name: name, description: description, archived: archived, creationTime: new Date(), masterkey: masterkey, iterations: iterations, salt: salt, authPublicKey: signPubKey, authPrivateKey: signPrvKey };
```
overheadhunter, line 328: I believe the order of arguments/fields of all DTOs should be the same in front- and backend. In the long run, this allows easier side-by-side coding of both components.

overheadhunter (changes requested):

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
infeo: Returning 404 only for the GET request would make our API inconsistent: For an archived vault v1 and device d1 `GET /v1/keys/d1` returns 404, but at the same time, the API would return 409 (Conflict) on `PUT /v1/keys/d1`.

I'd would suggest to either keep everything as it is or also adjust the PUT endpoint.

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
infeo: We only have one other entity, which gets PUT'ed: Devices in the device resource. And there we only create devices (no update).

frontend/src/common/backend.ts
infeo: Changed the backend in https://github.com/cryptomator/hub/pull/202/commits/615a370cc41358812383effd0c245eac30abd54c

backend/src/test/java/org/cryptomator/hub/api/VaultResourceTest.java
infeo: Fixed in https://github.com/cryptomator/hub/pull/202/commits/1331ecc9215fbfaa25d9797188c9c3bd72db072d

backend/src/test/java/org/cryptomator/hub/api/VaultResourceTest.java
infeo: Fixed in https://github.com/cryptomator/hub/pull/202/commits/1331ecc9215fbfaa25d9797188c9c3bd72db072d

backend/src/test/java/org/cryptomator/hub/api/VaultResourceTest.java
infeo: Fixed in https://github.com/cryptomator/hub/pull/202/commits/1331ecc9215fbfaa25d9797188c9c3bd72db072d

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
infeo: After a discussion we setteled to the following:
1. Symmetry in the API is nice, but it is our backend, so this is not necessary, but...
2. there could be a race condition: You already unlocked the vault in the desktop app, but someone else archived it. Still you are going to register your device
3. PUT and GET for `/vaultId/keys/deviceId` will return 410 if the vault is archived

infeo: Fixed in https://github.com/cryptomator/hub/pull/202/commits/c885038d5be367e2243c417cbcbebe5772850757

overheadhunter (approved):
Some API descriptions can still be simplified, but this shouldn't block this PR. I would rather do some separate documentation cleanup PR.

PR #203, opened 2023-06-07
Add compatibility for mobile apps + user profile page

### LINKED ISSUE #187: Add link for user to change password and configure 2FA

### Summary

Enable a non-admin user to change the password and configure 2FA

### Motivation

Currently, the only option is to trigger a password change and to configure 2FA in keycloak by an admin. This should be possible as a self-service for each user. The link to the corresponding user account page (https://login1.cryptomator.cloud/realms/[tenant]/account/) should be accessible via the menu on the top right.

### Considered Alternatives

None.

### Anything else?

_No response_

tobihagemann:
The following changes have been done to add compatibility for Cryptomator's mobile apps:

1. Added `type` to the Device entity and made sure that `DESKTOP` is default.
2. Added custom HTTP header to unlock response so that mobile apps can distinguish between free/paid licenses.

But as you can see, there is still this temporary code:

```java
return Response.ok(access.jwe).header("Cryptomator-Mobile-Access", license.isValid() ? "READ_WRITE" : "READ_ONLY").build();
```

I'm not sure what to name this header and what it should represent yet. E.g., if it's set to `READ_ONLY` but the user has bought the full version of the app, they should still have read/write access. That could be confusing. A different approach would be to have a header named `Hub-License-Edition` with `COMMUNITY` and `STANDARD` or something like that.

Furthermore, I've copied other useful changes from the `feature/refactored-access-grant` branch:

- Added user profile page, which combines the device list and user settings. (Fixes #187)
- Considered `BROWSER` device type already, even though it's not used yet.


backend/src/main/java/org/cryptomator/hub/api/DeviceResource.java
```diff
@@ -104,6 +104,7 @@ public Response remove(@PathParam("deviceId") @ValidId String deviceId) {
 
 	public record DeviceDto(@JsonProperty("id") @ValidId String id,
 							@JsonProperty("name") @NoHtmlOrScriptChars @NotBlank String name,
+							@JsonProperty("type") Device.Type type,
```
overheadhunter, line 107: Is there maybe an annotation for a default value, so we can get rid of the ternary operator?

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
```diff
@@ -259,7 +259,7 @@ public String unlock(@PathParam("vaultId") UUID vaultId, @PathParam("deviceId")
 		var access = AccessToken.unlock(vaultId, deviceId, jwt.getSubject());
 		if (access != null) {
 			UnlockVaultEvent.log(jwt.getSubject(), vaultId, deviceId, UnlockVaultEvent.Result.SUCCESS);
-			return access.jwe;
+			return Response.ok(access.jwe).header("Cryptomator-Mobile-Access", license.isValid() ? "READ_WRITE" : "READ_ONLY").build();
```
overheadhunter, line 262: Can we even reach this point (unlock: status 200), if the license is not valid?

backend/src/main/java/org/cryptomator/hub/license/LicenseHolder.java
```diff
@@ -67,6 +67,17 @@ public DecodedJWT get() {
 		return license;
 	}
 
+	/**
+	 * Checks if the license is valid.
+	 *
+	 * @return {@code true}, if the license _is set and valid_. Otherwise false.
+	 */
+	public boolean isValid() {
```
overheadhunter, line 75: bad name, since "valid" is too broad for just checking its validity period. How about negating the result and calling it `isExpired`?

overheadhunter (changes requested):
Adjustments in flyway test data are missing, aren't they?

backend/src/main/java/org/cryptomator/hub/api/DeviceResource.java
tobihagemann: Haven't found one. Tried `@JsonProperty(value = "type", defaultValue = "DESKTOP")` and also `@JsonEnumDefaultValue` on the `DESKTOP` enum value but they didn't work.

backend/src/main/java/org/cryptomator/hub/license/LicenseHolder.java
infeo: Why not splitting it into two functions then? A tad more code, but the meaning is clear.
* `boolean hasLicense()`
* `boolean isExpired()`

tobihagemann: Ended up with `isSet()`, see 325d050656758bd0659785e74fa4d4949a9ce85b.

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
tobihagemann: Your remark made me think about it again and I guess my answer also made clear what the only distinction is: Check if there is a license set or not, see 325d050656758bd0659785e74fa4d4949a9ce85b. Checking validity is already done by the `ActiveLicense` filter.

infeo: Regarding naming: Instead of naming the header after what it is used for, should we name it after its reason? Something like `Subscription-State` with values `ACTIVE` and `INACTIVE`

tobihagemann: Yeah, that's what I was thinking about as well but couldn't come up with a good name. I think I like `Hub-Subscription-State`! Made the changes in de5be39c2d37192c78eca4af69fde1d88306953e.

infeo:
For documentation: We talked about the custom header and if it can be faked. An idea is to sign the header with the vault key or the vault owner, in order to ensure that this is not possible.

tobihagemann:
Update on faking the custom header: It's impossible to sign the subscription state at the moment because this information is only known to the backend. We would need to introduce a new key pair for Hub and this is out of scope for this PR. However, we make sure that the mobile clients have to explictily check for the `ACTIVE` and `INACTIVE` values. When we introduce a signed subscription state at some point and use a different value, old mobile clients should treat it as `INACTIVE`.

infeo (approved):

overheadhunter (approved):
LGTM, I trust you have tried adding an "old" Cryptomator desktop client to Hub to check whether the "default" value works?

PR #204, opened 2023-06-19
Fix 204 without description.

chenkins:
Context:
```
org.openapitools.codegen.SpecValidationException: There were issues with the specification. The option can be disabled via validateSpec (Maven/Gradle) or --skip-validate-spec (CLI).
 | Error count: 1, Warning count: 0
Errors: 
	-attribute paths.'/api/billing/token'(put).responses.204.description is missing
```


infeo (approved):
Thanks for your contribution, @chenkins!

tobihagemann:
Hm, I wanted to have this change in `develop` as well but I guess, we have to cherry-pick this change then (or just wait until this gets merged to `develop`). Changing base didn't do what I expected. 😂

Issue #191, opened 2023-04-25
Verify Trust Level During Access Grant


overheadhunter:
### Please agree to the following

- [X] I have searched [existing issues](https://github.com/cryptomator/hub/issues?q=) for duplicates
- [X] I agree to follow this project's [Code of Conduct](https://github.com/cryptomator/hub/blob/develop/.github/CODE_OF_CONDUCT.md)

### Summary

Visualize trust level of users when granting access

### Motivation

As a vault owner, I want to be sure not to accidentally grant access to some untrusted user. Once we implemented #190, it is possible to show different trust levels and display a warning if trying to grant access to an unknown party.

### Anything else?

Trust levels could be:

1. _ultimate_, if a public key is signed by the vault owner herself
2. _good_, if there is a chain of signatures from the given public key leading to vault owner
3. _low_, otherwise

SailReal:
@overheadhunter this is also somehow implemented in https://github.com/cryptomator/hub/pull/281, should we close this issue as well, or keep it open for UI enhancements?

overheadhunter:
we decided to do further UI enhancements on demand

Issue #190, opened 2023-04-25
Sign User's Keys


overheadhunter:
### Please agree to the following

- [X] I have searched [existing issues](https://github.com/cryptomator/hub/issues?q=) for duplicates
- [X] I agree to follow this project's [Code of Conduct](https://github.com/cryptomator/hub/blob/develop/.github/CODE_OF_CONDUCT.md)

### Summary

After verifying another user's public key, I want to sign it

### Motivation

When encrypting something using somebody's public key, one needs to be sure that this public key is authentic. In order to increase trust, we want a verification mechanism and the ability to sign public keys once verification completed. The signature can then be used at a later point to prevent accidentally granting access to some untrusted user.

```[tasklist]
### Tasks
- [x] Decide, how to verify a user's public key (prompt for some digits of its hash?)
```

SailReal:
@overheadhunter this is implemented in https://github.com/cryptomator/hub/pull/281, so we can close this issue, right?

PR #207, opened 2023-07-06
refactored access management

### LINKED ISSUE #189: Self-Manage Devices

### Summary

I want to manage my devices without the need of an vault owner to re-grant access

### Motivation

As an unprivileged user I want to add and remove my own devices and instantly have access to vaults that I have been granted access to. I don't want to wait for the vault owner to rerun the "grant access" procedure.

### Considered Alternatives

_No response_

### Anything else?

_No response_

overheadhunter:
This is the first part of a series of PRs, which simplify access management.

## Introduction of User Key Pairs
We formerly only had _Device Key Pairs_. Vault keys have been shared with devices directly. This meant that the _Vault Owner_ who is in possession of the vault keys had to re-run the key distribution (via the "Update Permissions") button, every time a new device required access to the vault.

Now, we added _User Key Pairs_. Vault keys are now shared with the user. Each user will self-manage his or her devices. The device key pair does still exist and is required to decrypt the user's private key.

## Updated Entity Model

ERM

## Unlock Procedure
Due to the introduction of an intermediary key, an unlock operation requires to
1. first load the encrypted vault key (which can be decrypted using the user's private key) from `/api/vaults/{vaultId}/access-token` (response body is a ECDH-ES JWE)
2. then load the encrypted private key of the user (which can be decrypted using the device's private key) from `/api/devices/{deviceId}` (response body contains a JSON. attribute `userPrivateKey` contains ECDH-ES JWE)

Unlock Workflow

We still keep the legacy API endpoint (`/api/vaults/{vaultId}/keys/{deviceId}`) for compatibility reasons for a while. However it will only work for existing data. Any newly registered device can only be unlocked using the new workflow.

## Device Management
On the profile page, users can see their own devices. In order to add a new device, a _Setup Code_ (here called "Personal Hub Secret", tbd) is required. This code is a secret required to unwrap the user's private key in order to re-wrap it using the device's public key. Once the device is set up, its private key will allow to access the user's private key.

The _Setup Code_ can be access and changed any time by any fully set up device.

Device List

This fixes #189

## First Device
The browser used to log in to Hub is now also considered a device and a _Device Key Pair_ is generated and stored by the browser as well. The user will be prompted to name the browser accordingly (for identifying the browser, if they decide to revoke access from it later).

<p align="center">
  Initial Setup
</p>

The first device from which a login succeeds will also be responsible for generating the _User Key Pair_ together with the _Setup Code_.

The user's private key is then encrypted in JWE format with a) the device's public key (using an ECDH-ES) and b) using the setup code (using PBES2). The setup code in return is encrypted as an JWE using the user's private key (ECDH-ES again). This ensures that any authenticated device has access to the user's key pair and can thus decrypt the setup code again. Also, the setup code can be used to decrypt the user's key pair when setting up a new device.

## Vault Ownership
We used to use the _Vault Admin Password_ to protect a vault from being edited by anyone. Instead we will now use role-based permissions. The creator of a vault is the initial owner. Further owners can be added at a later time. Any owner has full administrative access to a vault (e.g. for adding members).

The _Vault Admin Password_ will be removed in a follow-up PR (as this one is already bursting).

Vault List

## Other
* `Device.publicKey` is now base64-encoded (used to be base64url for some reason)
* `Device.user_privateKey` is now non-null, i.e. if a device exists, we can be sure it is able to access the user's key pair

---

# Open Questions

- [ ] Currently, we generate v4 UUID (i.e. 122 bit randomness) as a _Setup Code_, which is less than the key it protects. In my opinion this is sufficient, this setup code is then fed into PBKDF2 (currently defaulting to [1 mio iterations](https://github.com/cryptomator/hub/blob/93778ee85433f5a0ddda634d36dcbed2a941c55e/frontend/src/common/jwe.ts#L246)) to derive the key used to wrap the user's private key.

# Next Steps

- [x] ~Remove _Vault Admin Password_~ see #209
- [x] ~Add _Claim Vault Ownership_ button (for migrating old data that still uses the vault admin password)~ see #210
- [x] ~WoT for user keys ("verify users")~ see #190
- [x] ~Warning in "grant access" dialog, if granting access to unverified users~ see #191
- [x] ~Add _Reset User Account_ feature (re-roll user key pair, keep devices, losing key signatures and vault access)~ see #208
- [x] Remove _Legacy Devices_ during re-registration as a _Device_ using the updated API


backend/src/main/resources/application.properties
```diff
@@ -60,6 +60,7 @@ quarkus.datasource.jdbc.max-size=16
 quarkus.hibernate-orm.dialect=org.hibernate.dialect.PostgreSQL10Dialect
 quarkus.hibernate-orm.database.globally-quoted-identifiers=true
 quarkus.flyway.migrate-at-start=true
+quarkus.flyway.ignore-missing-migrations=true
```
tobihagemann, line 63: What's that for?

frontend/src/components/GrantPermissionDialog.vue
```diff
@@ -96,11 +96,11 @@ async function grantAccess() {
...
 }
 
-async function giveDevicesAccess(devices: DeviceDto[]) {
-  for (const device of devices) {
-    const publicKey = base64url.parse(device.publicKey);
-    const jwe = await props.vaultKeys.encryptForDevice(publicKey);
-    await backend.vaults.grantAccess(props.vault.id, device.id, jwe, props.vaultKeys);
+async function giveUsersAccess(users: UserDto[]) {
+  for (const user of users) {
+    const publicKey = base64.parse(user.publicKey!); // TODO: can we be sure the user has a public key?
```
tobihagemann, line 101: Does `User.requiringAccessGrant` only return users with a public key?

frontend/src/components/RegenerateRecoveryCodeDialog.vue
```diff
@@ -0,0 +1,185 @@
...
+    processing.value = false;
+  }
+}
+
+async function copyRecoveryCode() {
+  await navigator.clipboard.writeText(recoveryCode.value);
+  copiedRecoveryCode.value = true;
+  debouncedCopyFinish();
+}
+</script>
```
tobihagemann, line 185: Should we rename the file and replace all occurrences of recovery code with setup code?

tobihagemann (changes requested):

backend/src/main/resources/application.properties
overheadhunter: Please don't ask me questions that I have no answer to. 😅

I probably just forgot to revert a local change that I needed for testing.

overheadhunter: This is a workaround for a hot reloading issue during `quarkus:dev`. See #212.

I've committed a workaround on develop, so this should no longer be a part of this PR. (diff removed in 57ce308)

frontend/src/components/GrantPermissionDialog.vue
overheadhunter: fixed in bf4a625

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
```diff
@@ -236,81 +252,109 @@ private Response removeAuthority(UUID vaultId, String authorityId) {
...
+		}
+
+		var access = AccessToken.unlock(vaultId, jwt.getSubject());
+		if (access != null) {
+			return access.vaultKey;
+		} else if (Vault.findById(vaultId) == null) {
+			throw new NotFoundException("No such vault.");
+		} else {
+			throw new ForbiddenException("Access to this vault not granted.");
+		}
```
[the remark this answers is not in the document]
overheadhunter: `device_id` is `NOT NULL`... Not exactly compatible with this refactoring. I see two options:

1. Remove this column in this PR
2. Don't add the column in the first place in 1.2.0. It won't last very long anyways 😉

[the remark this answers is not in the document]
tobihagemann: Oh... I've looked into the flowchart again. How about the following:

1. In `/api/vaults/{vaultId}/access-token`, we only add the audit event log for the last case, meaning if the vault unlock failed due to missing vault access.
2. In `/api/devices/{deviceId}`, we add the audit event log for a successful unlock. But I wonder, why did we remove the case `NO_SUCH_DEVICE` then?

I know that these APIs can also be called in reverse but you probably have this "problem" in any order. Even in this resource, the first case doesn't have to lead to a successful unlock because the user might still have to register their device.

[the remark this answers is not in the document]
overheadhunter: The latter does not need to be an unlock. There can be plenty reasons, why a client GETs device info.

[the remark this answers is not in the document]
overheadhunter: added in e9411b4

tobihagemann (approved):

PR #211, opened 2023-07-10
Deprecate vault admin password

### LINKED ISSUE #209: Deprecate `Vault Admin Password`

### Summary

Switch back to role-based access control in access-restricted APIs

### Motivation

Since we needed a mechanism to decrypt certain vault-specific encapsulated keys in the browser, we relied on the _Vault Admin Password_. **This has now been obsoleted** by the fact that with #207 browsers are promoted to being _Devices_ which can access the user's key pair. So if said vault-specific keys are now encapsulated for authorized users, there is no need for the _Vault Admin Password_ any longer.

Among those keys were not just the vault's masterkey, but also a signing key (added in #81 and #87) for signing http requests that were only allowed to be made by vault owners (e.g. managing vault members). The backend relied on these signatures to verify whether the user is allowed to make such a request.

Since #207 re-introduces vault roles, the backend can rely on those again.

### Considered Alternatives

_No response_

### Anything else?

We can not completely remove the _Vault Admin Password_ yet, as we need to migrate existing data first.
```[tasklist]
### Tasks
- [x] Replace `@VaultAdminOnlyFilter`
- [x] Remove _Vault Admin Password_ prompt from Hub frontend
- [x] Display "Manage Vault" button depending on role
- [ ] https://github.com/cryptomator/hub/issues/214
```

### LINKED ISSUE #210: Migrate former "Vault Admin" to `OWNER` role

### Summary

Proof ownership of existing vaults by entering the "Vault Admin Password"

### Motivation

After #81 and #87 the _vault owner_ has been removed and got replaced by the _Vault Admin Password_, which served as a means to proof authorization of vault-related administrative actions. With #207 we reintroduced the `OWNER` role and paved the way of removing the _Vault Admin Password_, which is mostly done in #209.

However, for existing Hub installations, we need a way to decide, which user becomes an `OWNER` of which vault. Since users used to show proof by entering the _Vault Admin Password_, we decided to rely on this.

So non-`OWNER` users should be shown a "Claim Ownership" button, which triggers a prompt asking for the _Vault Admin Password_. If entered correctly, a request is made to the backend, containing a random challenge and a response signed with the deprecated `authenticationPrivateKey`. The backend will verify the signature using the `authenticationPublicKey` and on success grants the `OWNER` role to the current user.

### Considered Alternatives

_No response_

### Anything else?

_No response_

overheadhunter:
This fixes #209:

`@VaultAdminOnlyFilter` got replaced by `@VaultRole(VaultAccess.Role.OWNER)`.

The UI will no longer ask for the _Vault Admin Password_ when a user is the vault's owner.

```[tasklist]
### Tasks
- [x] Refresh vault list after claiming ownership
- [ ] #214
```


tobihagemann (changes requested):

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
```diff
@@ -444,15 +441,73 @@ public Response createOrUpdate(@PathParam("vaultId") UUID vaultId, @Valid @NotNu
...
+		}
+
+		vault.salt = null;
+		vault.iterations = null;
+		vault.masterkey = null;
+		vault.authenticationPrivateKey = null;
+		vault.authenticationPublicKey = null;
+		vault.persist();
+
+		// TODO: log some event?
```
tobihagemann, line 497: I think it would make sense to introduce a new audit event. "Update Vault" doesn't really reflect this important change. How should we call this? "Deprecate Vault Admin Password"?

tobihagemann (changes requested):

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
overheadhunter: Claimed ownership?

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
```diff
@@ -374,6 +373,7 @@ public Response grantAccess(@PathParam("vaultId") UUID vaultId, @PathParam("user
 	@GET
 	@Path("/{vaultId}")
 	@RolesAllowed("user")
+	// @VaultRole(VaultAccess.Role.MEMBER) // TODO: members and admin may do this...
```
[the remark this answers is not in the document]
overheadhunter: No, hence the TODO 😉

I think we should use separate methods for admin and members.

In fact I believe for the admin interface the frontend really just makes use of `getAll`...?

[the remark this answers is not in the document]
tobihagemann: This method is used in the frontend when retrieving data for the "vault details" view. An admin is able to list all vaults, regardless of membership. This means that this method is indeed called by admins _and_ members.

In this case, I think separate methods would be better. But that would require some if/else logic in the frontend, though. Maybe it's better this way because admins (who are not members) might need a different view for things like "claim ownership via recovery key" anyway (if the vault somehow has no owner anymore).

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
tobihagemann: I already introduced an "Update Vault Member" audit event that shows which user claimed ownership. This log event is for "nulling" the vault admin password fields, not really for claiming ownership. That's why "Update Vault" would fit better, but... it wouldn't then tell anyone anything about the "removal" of the vault admin password.

tobihagemann: I'm leaning towards "Remove Vault Admin Password".

overheadhunter: That is a rather technical perspective. From the user's point of view he never removed a password, he pressed "claim ownership".
Thus I wouldn't split this into two events.

tobihagemann: Makes sense. It may confuse admins when the audit event doesn't match the action from a user perspective. Then I'll introduce a new "Claim Vault Ownership" audit event.

Regarding "two" events, I was just filling in the TODOs, to be honest. Shall I delete `AuditEventVaultMemberUpdate.log()`/`AuditEventVaultMemberAdd.log()` in the lines before that?

overheadhunter: Not sure, we have these events during vault creation as well. So maybe "two" is not the problem, as long as the events can be related via time/actor.

tobihagemann: Fixed in 92b4fb35ebabc48f61baa705122e78cc23a37259.

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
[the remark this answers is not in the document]
overheadhunter: Maybe we need some annotation like this, but this exceeds the scope of this PR:

```java
@AuthRequireEither({
    @RolesAllowed("admin"),
    @AuthRequireAll({
        @RolesAllowed("user"),
        @VaultRole(VaultAccess.Role.MEMBER)
    })
})
```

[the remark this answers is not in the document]
tobihagemann: All right, let's keep this TODO then and don't blow up this PR.

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
```diff
@@ -116,22 +123,21 @@ public List<VaultDto> getAllVaults() {
...
-	@APIResponse(responseCode = "403", description = "VaultAdminAuthorizationJWT expired or not yet valid")
-	@APIResponse(responseCode = "404", description = "vault not found")
-	public List<AuthorityDto> getMembers(@PathParam("vaultId") UUID vaultId) {
-		var vault = Vault.<Vault>findByIdOptional(vaultId).orElseThrow(NotFoundException::new);
+	@APIResponse(responseCode = "403", description = "not a vault owner")
+	public List<MemberDto> getMembers(@PathParam("vaultId") UUID vaultId) {
+		var vault = Vault.<Vault>findById(vaultId); // should always be found, since @VaultRole filter would have triggered
 
 		return vault.directMembers.stream().map(authority -> {
+			VaultAccess access = VaultAccess.findById(new VaultAccess.Id(vaultId, authority.id)); // TODO: inefficient, would be better if role is already part of the query result
```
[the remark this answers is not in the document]
tobihagemann: Maybe some more info on this: When listing all vault members, the frontend needs the vault role of each member. The vault entity already has a "direct members" array, but the "authority" entities don't contain a vault role. That's why the vault role has to be determined for _all_ authorities separately.

That's why I claimed that this is an inefficient approach (`O(n)`).
- Can we live with that or should we refactor it to something like `VaultAccess.list()` (not sure how this exactly works but basically "find by multiple IDs").
- Or should the vault role be part of the authority entity? Maybe as an optional field? Not particularly a fan of this.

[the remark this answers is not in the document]
overheadhunter: I'll take a closer look at this. There are certainly better ways to do this.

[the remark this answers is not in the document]
overheadhunter: fixed in d0d4288. Note that this query will now _always_ fetch vault + authority eagerly. Alternatively we could use `javax.persistence.fetchgraph` with an appropriate object graph as a query hint.

tobihagemann (approved):

PR #213, opened 2023-07-13
Update Audit Events

tobihagemann:
At first, my goal was to update the `CreateVaultEvent` with name/description and add a new `UpdateVaultEvent`.

But during testing, I felt that the audit log still felt incomplete. That's why I also added `GrantVaultAccessEvent`, `RegisterDeviceEvent`, and `RemoveDeviceEvent`.

Screenshot 2023-07-18 at 16 49 41


backend/src/main/java/org/cryptomator/hub/entities/GrantVaultAccessEvent.java
```diff
@@ -0,0 +1,53 @@
...
+	public static final String TYPE = "GRANT_VAULT_ACCESS";
+
+	@Column(name = "user_id")
+	public String userId;
+
+	@Column(name = "vault_id")
+	public UUID vaultId;
+
+	@Column(name = "authority_id")
+	public String authorityId;
```
overheadhunter, line 26: Regarding `userId` and `authorityId` it is unclear who granted access and whom was granted access.

Maybe we can find better names.

backend/src/main/resources/org/cryptomator/hub/flyway/V11__Update_Audit_Events.sql
```diff
@@ -0,0 +1,50 @@
+-- Delete all audit events so that we don't have to deal with migration of old events
+-- This is fine because audit logging hasn't been introduced in a stable release yet
+TRUNCATE "audit_event" CASCADE;
```
overheadhunter, line 3: `TRUNCATE`

This is fine. 🐶🔥

frontend/src/common/backend.ts
```diff
@@ -169,12 +169,32 @@ export type VersionDto = {
 export type AuditEventDto = {
   id: number;
   timestamp: Date;
-  type: 'CREATE_VAULT' | 'UNLOCK_VAULT' | 'UPDATE_VAULT_MEMBERSHIP';
+  type: 'CREATE_VAULT' | 'GRANT_VAULT_ACCESS' | 'REGISTER_DEVICE' | 'REMOVE_DEVICE' | 'UNLOCK_VAULT' | 'UPDATE_VAULT' | 'UPDATE_VAULT_MEMBERSHIP';
```
overheadhunter, line 172: With the steadily growing number of events, we should move related code to a separate file.

Maybe we can (later!) get rid of `backend.ts` altogether, cutting by domain, not by layer.

backend/src/main/resources/org/cryptomator/hub/flyway/V11__Update_Audit_Events.sql
tobihagemann: I hope it is? 😂 It sounded great!

https://stackoverflow.com/questions/13223820/postgresql-delete-all-content
https://www.postgresql.org/docs/current/sql-truncate.html

backend/src/main/java/org/cryptomator/hub/entities/GrantVaultAccessEvent.java
tobihagemann: Yeah, I just used the same naming pattern from other events. In this case, it's similar to `UpdateVaultMembershipEvent`. And also from all the other events because they all have a `user_id` to identify the user that caused this event. This could indeed be improved.

But should we leave `user_id` as it is? I don't think that the name is _that_ bad. Or should we refactor it to something new like `actor_id` or `created_by` (which doesn't really fit with the other columns like `timestamp`)? And I wonder if this column is a common one for _all_ events, then we could move it to `AuditEvent`.

Regarding `authority_id`: How about `granted_authority_id` for grant access? And for update vault membership, `updated_authority_id`?

tobihagemann:
Now I'm wondering about `UpdateVaultMembershipEvent` (with its `operation` column for `ADD` and `REMOVE`) vs. `RegisterDeviceEvent` + `RemoveDeviceEvent`.

Should we split up `UpdateVaultMembershipEvent` as well? E.g., to `AddVaultMembershipEvent` and `RemoveVaultMembershipEvent`?

backend/src/main/java/org/cryptomator/hub/api/AuditLogResource.java
```diff
@@ -56,10 +61,11 @@ public List<AuditEventDto> getAllEvents(@QueryParam("startDate") Instant startDa
...
 	@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type")
-	@JsonSubTypes({
-			@JsonSubTypes.Type(value = CreateVaultEventDto.class, name = CreateVaultEvent.TYPE),
-			@JsonSubTypes.Type(value = UnlockVaultEventDto.class, name = UnlockVaultEvent.TYPE),
-			@JsonSubTypes.Type(value = UpdateVaultMembershipEventDto.class, name = UpdateVaultMembershipEvent.TYPE)
+	@JsonSubTypes({ //
+			@JsonSubTypes.Type(value = CreateVaultEventDto.class, name = CreateVaultEvent.TYPE), //
+			@JsonSubTypes.Type(value = UnlockVaultEventDto.class, name = UnlockVaultEvent.TYPE), //
+			@JsonSubTypes.Type(value = UpdateVaultEventDto.class, name = UpdateVaultEvent.TYPE), //
+			@JsonSubTypes.Type(value = UpdateVaultMembershipEventDto.class, name = UpdateVaultMembershipEvent.TYPE) //
```
infeo, line 68: Here the Subtypes for `RegisterDeviceEvent`, `RemoveDeviceEvent` and `GrantVaultAccessEvent` are missing

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
```diff
@@ -362,13 +365,14 @@ public Response createOrUpdate(@PathParam("vaultId") UUID vaultId, @Valid @NotNu
 		//update new or existing vault
 		vault.name = vaultDto.name;
 		vault.description = vaultDto.description;
-		vault.archived = vaultDto.archived;
+		vault.archived = isCreated ? false : vaultDto.archived;
```
infeo, line 368: Although understandable from user perspective, this actually changes the logic and must be noted in the `@Operation( ... description="")` field and we should also cover this in a unit test.

@overheadhunter From the backend view, it does not matter if we create directly an archived vault. What do you think?

backend/src/main/java/org/cryptomator/hub/entities/GrantVaultAccessEvent.java
infeo: >I don't think that the name is that bad

I second that. If you give it one though, you notice, that only an actual user can create an event, hence `user_id` is always the cause while authority_id must be part of the target.

infeo (changes requested):
When this PR leaves draft status, can you also add a screenshot of the frontend containing some of the new/altered events? Doesn't have to be a complete view, but to see the UI change.

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
tobihagemann: Another reason is that I didn't want to include an `archived` column in `CreateVaultEvent`. So I came to the conclusion that it wouldn't be that useful to create an archived vault.

backend/src/main/java/org/cryptomator/hub/api/AuditLogResource.java
tobihagemann: Fixed in a3fec58ae9dfa2c41a776279e8ae583072a4b1b1.

backend/src/main/java/org/cryptomator/hub/entities/GrantVaultAccessEvent.java
tobihagemann: > And I wonder if this column is a common one for _all_ events, then we could move it to `AuditEvent`.

One reason against this is that if we add "system" events in the future that are not caused by a logged-in user.

tobihagemann:
Screenshot 2023-07-14 at 17 53 49

backend/src/main/java/org/cryptomator/hub/entities/GrantVaultAccessEvent.java
infeo: Let's change it, while we still have the chance.

I like `actor_id` and then naming the columns of the target depending on the event. Another option would be to follow the JWT standard with `issuer`( == user_id) and `subject` (==authority_id).

backend/src/test/resources/org/cryptomator/hub/flyway/V9999__Test_Data.sql
```diff
@@ -65,22 +65,68 @@ VALUES
...
 INSERT INTO "audit_event" ("id", "timestamp", "type")
 VALUES
-    (999, '1970-01-01T00:00:02.900Z', 'CREATE_VAULT'),
-    (1000, '1970-01-01T00:00:03Z', 'UPDATE_VAULT_MEMBERSHIP'),
-    (1001, '1970-01-01T00:00:04Z', 'UPDATE_VAULT_MEMBERSHIP'),
-    (1111, '1970-01-01T00:00:04.999Z', 'UNLOCK_VAULT'),
-    (4242, '1970-01-01T00:00:05Z', 'UNLOCK_VAULT');
+    (10, '2020-02-20T20:20:20.010Z', 'CREATE_VAULT'),
+    (11, '2020-02-20T20:20:20.011Z', 'UPDATE_VAULT_MEMBERSHIP'),
+    (12, '2020-02-20T20:20:20.012Z', 'UPDATE_VAULT_MEMBERSHIP'),
```
infeo, line 70: The reason to directly start with quite high numbers is, that also preceding tests generate audit events and they might then fail, because an audit event with sequence number 10 already exists. My suggestion is to directly start at 1000, or at least 100.

infeo (changes requested):
Regarding splitting up `UpdateVaultMemberShip`: Splitting up means more boilerplate code (two tables classes, each with similar entries, etc) in the backend. Also, the more tables we have, the more JOINS need to be made when querying audit events.

@tobihagemann Does the splitting give some benefits in the frontend? Like filtering? Note that for specififcally filter to "archived" vault events, the filter function alrady needs some complexity.

backend/src/test/resources/org/cryptomator/hub/flyway/V9999__Test_Data.sql
tobihagemann: That's why I added this 16915815d10b17ce8bbb5464c39a6083dcf9d830.

backend/src/main/java/org/cryptomator/hub/entities/GrantVaultAccessEvent.java
tobihagemann: Fixed in 01b3888d7c19313a85bae37f4a1fb06421682198 by renaming `user_id` to be more specific, e.g., `granted_by` or `created_by`. I've left `authority_id` as is because it felt clear to me that this was the subject (as in other events that a vault or device is the subject).

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
tobihagemann: Fixed in 8b825b4fd9a1c4fae9c859d447b0218af055bb1e.

tobihagemann:
@infeo I've split up `UpdateVaultMembership` into their own add and remove events after discussing this with @overheadhunter.

backend/src/main/java/org/cryptomator/hub/api/AuditLogResource.java
```diff
@@ -79,51 +81,57 @@ public interface AuditEventDto {
...
-				return new UnlockVaultEventDto(uve.id, uve.timestamp, UnlockVaultEvent.TYPE, uve.userId, uve.vaultId, uve.deviceId, uve.result);
+				return new UnlockVaultEventDto(uve.id, uve.timestamp, UnlockVaultEvent.TYPE, uve.unlockedBy, uve.vaultId, uve.deviceId, uve.result);
 			} else if (entity instanceof UpdateVaultEvent uve) {
-				return new UpdateVaultEventDto(uve.id, uve.timestamp, UpdateVaultEvent.TYPE, uve.userId, uve.vaultId, uve.vaultName, uve.vaultDescription, uve.vaultArchived);
-			} else if (entity instanceof UpdateVaultMembershipEvent uvme) {
-				return new UpdateVaultMembershipEventDto(uvme.id, uvme.timestamp, UpdateVaultMembershipEvent.TYPE, uvme.userId, uvme.vaultId, uvme.authorityId, uvme.operation);
+				return new UpdateVaultEventDto(uve.id, uve.timestamp, UpdateVaultEvent.TYPE, uve.updatedBy, uve.vaultId, uve.vaultName, uve.vaultDescription, uve.vaultArchived);
 			} else {
 				throw new UnsupportedOperationException("conversion not implemented for event type " + entity.getClass());
 			}
```
overheadhunter, line 102: a `switch` might look better 😉

backend/src/main/resources/org/cryptomator/hub/flyway/V11__Update_Audit_Events.sql
```diff
@@ -3,14 +3,41 @@
 TRUNCATE "audit_event" CASCADE;
 ALTER SEQUENCE "audit_event_id_seq" RESTART;
 
+-- Add and rename columns to/of existing events
```
overheadhunter, line 6: Let's go nuclear and promote the `TRUNCATE` to a `DROP` and use this chance to rename tables, so they have a common prefix `audit_event_*`. What do you think?

overheadhunter (changes requested):

backend/src/main/java/org/cryptomator/hub/entities/RemoveVaultMembershipEvent.java
```diff
@@ -0,0 +1,53 @@
...
+import jakarta.persistence.Table;
+
+import java.time.Instant;
+import java.util.Objects;
+import java.util.UUID;
+
+@Entity
+@Table(name = "remove_vault_membership_event")
+@DiscriminatorValue(RemoveVaultMembershipEvent.TYPE)
+public class RemoveVaultMembershipEvent extends AuditEvent {
```
infeo, line 15: Looking at the wording, this could be imroved by renaming it to `RemoveVaultMemberEvent` now. Similarly, `AddVaultMembershipEvent`  to  `AddVaultMemberEvent`.

backend/src/test/java/org/cryptomator/hub/api/VaultResourceTest.java
```diff
@@ -843,6 +843,25 @@ public void testCreateVault2() {
 					.then().statusCode(400);
 		}
 
+		@Test
+		@DisplayName("PUT /vaults/7E57C0DE-0000-4000-8000-0001FFFFAAAA returns 201")
```
infeo, line 847: Can we add some more text here? E.g. `...returns 201 ignoring archived flag`

backend/src/main/java/org/cryptomator/hub/entities/RemoveVaultMembershipEvent.java
infeo: I just noticed, this would trigger renamings at some places (database, frontend). If this is a major blocker, say it and we ignore this request.

infeo (changes requested):

backend/src/main/resources/org/cryptomator/hub/flyway/V11__Update_Audit_Events.sql
tobihagemann: I don't think it's necessary to promote `TRUNCATE` to `DROP` because we'd probably leave the "parent" table `audit_event` as is, right?

But I think it makes sense to rename the tables to `audit_event_*`, e.g., `audit_event_unlock_vault`, so that the audit event tables are grouped together. Or maybe even `audit_event_vault_unlock` so that "audit events" of "vault" are grouped together (`*_create`, `*_update`, `*_add_member`, `*_remove_member`, `*_grant_access`).

backend/src/main/java/org/cryptomator/hub/api/AuditLogResource.java
tobihagemann: Added TODO in 0df8420076f892e01e91ddb372e795ac40ee0c63 because I didn't want to introduce a new compilation flag in this PR.

backend/src/main/resources/org/cryptomator/hub/flyway/V11__Update_Audit_Events.sql
tobihagemann: Fixed in 0df8420076f892e01e91ddb372e795ac40ee0c63.

backend/src/main/java/org/cryptomator/hub/entities/RemoveVaultMembershipEvent.java
tobihagemann: Fixed in 0df8420076f892e01e91ddb372e795ac40ee0c63.

backend/src/test/java/org/cryptomator/hub/api/VaultResourceTest.java
tobihagemann: Fixed in 0df8420076f892e01e91ddb372e795ac40ee0c63.

overheadhunter (approved):
I really just wanted to rename the tables, since there are no namespaces in SQL. But renaming the entities doesn't hurt either. So I'm fine with this :)

infeo (approved):
uff, that's a lotta renaming. But good, now everything is aligned.

PR #215, opened 2023-07-26
Update Audit Event "Unlock Vault" to "Retrieve Vault Key"

tobihagemann:
We noticed in https://github.com/cryptomator/hub/pull/207#pullrequestreview-1528786098 that we'd like to drop the `device_id` column in the "unlock vault" audit event because this information will no longer be available after the access management refactoring.

Also, I'd like to rename the "unlock vault" audit event to "retrieve vault key" to be more precise about what the audit event is actually logging.


backend/src/main/java/org/cryptomator/hub/api/AuditLogResource.java
```diff
@@ -127,6 +123,10 @@ record AuditEventVaultAccessGrantDto(long id, Instant timestamp, String type, @J
 										 @JsonProperty("authorityId") String authorityId) implements AuditEventDto {
 	}
 
+	record AuditEventVaultKeyRetrieveDto(long id, Instant timestamp, String type, @JsonProperty("retrievedBy") String retrievedBy, @JsonProperty("vaultId") UUID vaultId,
```
overheadhunter, line 126: Re naming, I would suggest either `RetrieveKey` or `KeyRetrieval`.

overheadhunter (changes requested):

backend/src/main/java/org/cryptomator/hub/api/AuditLogResource.java
overheadhunter: > But if this leads to confusion ...

Indeed, I read it like `AuditEvent-Vault-KeyRetrieve` 😅

For me, it is not `AuditEvent-Noun-Verb` but rather `AuditEvent-Domain-Event` with this belonging into the "Vault" domain/namespace.

Hence, the following would be my preferred naming scheme.

> * `AuditEvent-Vault-Create`
> * `AuditEvent-Vault-GrantAccess`
> * `AuditEvent-Vault-RetrieveKey`
> * `AuditEvent-Vault-AddMember`

That said, as already noted in https://github.com/cryptomator/hub/pull/213#pullrequestreview-1536702327, when I asked you to prefix the _table names_, I really didn't want to force the naming scheme upon all entities, DTOs and .ts types, since those would be namespaced by means of the used programming language anyway. 😉

overheadhunter (approved):
After discussing the naming scheme offline, we decided to keep it as is.

PR #228, opened 2023-10-24
Reset User Account

### LINKED ISSUE #208: Reset User Account
As a user, who lost access to all devices as well as the _Setup Code_, I want to reset my user account to gain access again.

Resetting will re-roll the user's key pair. I.e. while the user can remain a vault member, keys need to be re-shared ("Update Permissions" button).

Furthermore, we need to decide, whether the user needs to re-validate his devices, as in the meantime a malicious admin could have sneaked in fake devices which we must not allow to receive a copy of the user's new key.
```[tasklist]
### Tasks
- [x] discuss necessity of removing the user's devices
```

overheadhunter:
Fixes #208:

This deletes the user's key pair as well as devices. Furthermore all access tokens for the deleted key will be deleted.

### DIFF
--- a/backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
+++ b/backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
@@ -7,12 +7,14 @@ import jakarta.transaction.Transactional;
 import jakarta.validation.Valid;
 import jakarta.ws.rs.Consumes;
 import jakarta.ws.rs.GET;
+import jakarta.ws.rs.POST;
 import jakarta.ws.rs.PUT;
 import jakarta.ws.rs.Path;
 import jakarta.ws.rs.Produces;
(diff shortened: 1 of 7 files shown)


overheadhunter:
Note that deleting the browser key pair is probably dead code. If you think about the conditions when entering `/app/setup`:

|                       | browser key pair exists | browser key pair missing          |
|-----------------------|-------------------------|-----------------------------------|
| user key pair exists  | ✅ no setup required    | enter setup code, name new device |
| user key pair missing | ❌ strange state        | create new user key pair          |

Technically, it is possible for a browser key pair to still exist if the account has been reset on a different device. However, one the program logic would treat this the same as if the browser key did not exist and re-roll it anyway.

tobihagemann (approved):
I'm fine with the "dead code". 👍

PR #230, opened 2023-10-27
Fix unarchive vaults

SailReal:
To unarchive vaults again, it was required to 
* add an optional `ignoreArchivedFlag` query parameter to the `"/{vaultId}/access-token"` resource. If set to `true`, the archived state of the vault will be ignored when retrieving an access token. We need this because the frontend uses it to check if the user is actually the owner of the vault
* the database query `Vault.accessibleByUserAndRole` and `Vault.accessibleByUser` now also returns the archived vaults and the frontend excludes the archived vaults depending on what it wants to show to the user
* the non-admin user can now also filter the vault list between "Accessible" and "Owned by me" so that the user can unarchive their vaults

### DIFF
--- a/backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
+++ b/backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
@@ -310,11 +310,11 @@ public class VaultResource {
 	@APIResponse(responseCode = "402", description = "number of effective vault users exceeds available license seats")
 	@APIResponse(responseCode = "403", description = "not a vault member")
 	@APIResponse(responseCode = "404", description = "unknown vault")
-	@APIResponse(responseCode = "410", description = "Vault is archived")
+	@APIResponse(responseCode = "410", description = "Vault is archived. Only returned if ignoreArchivedVault is false or not set, otherwise the archived flag is ignored")
 	@ActiveLicense // may throw 402
-	public String unlock(@PathParam("vaultId") UUID vaultId) {
+	public String unlock(@PathParam("vaultId") UUID vaultId, @QueryParam("ignoreArchivedFlag") Optional<Boolean> ignoreArchivedFlag) {
 		var vault = Vault.<Vault>findById(vaultId); // should always be found, since @VaultRole filter would have triggered
-		if (vault.archived) {
+		if (vault.archived && !ignoreArchivedFlag.orElse(false)) {
 			throw new GoneException("Vault is archived.");
 		}
 
--- a/backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
+++ b/backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
@@ -86,7 +86,7 @@ public class VaultResource {
 	@RolesAllowed("user")
 	@Produces(MediaType.APPLICATION_JSON)
 	@Transactional
-	@Operation(summary = "list all accessible vaults", description = "list all (non-archived) vaults that have been shared with the currently logged in user or a group in wich this user is")
+	@Operation(summary = "list all accessible vaults", description = "list all vaults that have been shared with the currently logged in user or a group in wich this user is")
 	public List<VaultDto> getAccessible(@Nullable @QueryParam("role") VaultAccess.Role role) {
 		var currentUserId = jwt.getSubject();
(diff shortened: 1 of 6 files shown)


backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
```diff
@@ -310,11 +310,11 @@ public Response legacyUnlock(@PathParam("vaultId") UUID vaultId, @PathParam("dev
 	@APIResponse(responseCode = "402", description = "number of effective vault users exceeds available license seats")
 	@APIResponse(responseCode = "403", description = "not a vault member")
 	@APIResponse(responseCode = "404", description = "unknown vault")
-	@APIResponse(responseCode = "410", description = "Vault is archived")
+	@APIResponse(responseCode = "410", description = "Vault is archived. Only returned if ignoreArchivedVault is false or not set, otherwise the archived flag is ignored")
 	@ActiveLicense // may throw 402
-	public String unlock(@PathParam("vaultId") UUID vaultId) {
+	public String unlock(@PathParam("vaultId") UUID vaultId, @QueryParam("ignoreArchivedFlag") Optional<Boolean> ignoreArchivedFlag) {
```
overheadhunter, line 315: 1. Instead of an `Optional`, use a plain `@DefaultValue("false") boolean`
2. I'd suggest to use a more "fluent" name, such as `unlock [...] despiteArchived` or `evenIfArchived` in the exposed API (internal variable name can be different to match internal logic)

```suggestion
	public String unlock(@PathParam("vaultId") UUID vaultId, @QueryParam("evenIfArchived") @DefaultValue("false") ignoreArchived) {
```

frontend/src/common/backend.ts
```diff
@@ -257,7 +257,7 @@ class VaultService {
   }
 
   public async accessToken(vaultId: string): Promise<string> {
-    return axiosAuth.get(`/vaults/${vaultId}/access-token`, { headers: { 'Content-Type': 'text/plain' } })
+    return axiosAuth.get(`/vaults/${vaultId}/access-token?ignoreArchivedFlag=true`, { headers: { 'Content-Type': 'text/plain' } })
```
overheadhunter, line 260: Should this be hard-coded? Even if the frontend always needs to ignore this flag, I'd prefer to expose the parameter via the method signature (`accessToken(vaultId: string, evenIfArchived = false): Promise<string>`) and explicitly set this boolean to true from the component when needed.

frontend/src/components/VaultList.vue
```diff
@@ -138,11 +138,7 @@ const ownsSelectedVault = computed(() => {
 
 const isAdmin = ref<boolean>();
 
-const filterOptions = {
-  accessibleVaults: t('vaultList.filter.entry.accessibleVaults'),
-  ownedVaults: t('vaultList.filter.entry.ownedVaults'),
-  allVaults: t('vaultList.filter.entry.allVaults'),
-};
+const filterOptions = ref< {[key: string]: string} >({accessibleVaults: t('vaultList.filter.entry.accessibleVaults'), ownedVaults: t('vaultList.filter.entry.ownedVaults')});
```
overheadhunter, line 141: I like the multiline version better (shouldn't our linter enforce it?)

overheadhunter (changes requested):

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
SailReal: Applied in [e959779](https://github.com/cryptomator/hub/pull/230/commits/e959779b5f86fefe6739fa6db7baabc142e14c89)

frontend/src/common/backend.ts
SailReal: Applied in [ac513a6](https://github.com/cryptomator/hub/pull/230/commits/ac513a677e950ad490821a75a823408851332598)

frontend/src/components/VaultList.vue
SailReal: Applied in [fe12a83](https://github.com/cryptomator/hub/pull/230/commits/fe12a83fd2b435e14bc4d6fbcf277eccdbfb2265)

overheadhunter (approved):

PR #235, opened 2023-11-06
View users with their key fingerprint who should be granted permissions

SailReal:
image


tobihagemann (changes requested):
I'm totally fine with this PR. But we should add _some_ way for vault owners to verify these fingerprints with the users if they choose to.

As discussed, a user should be able to check their fingerprint. I believe the user profile would make most sense?

overheadhunter:
> I'm totally fine with this PR. But we should add _some_ way for vault owners to verify these fingerprints with the users if they choose to.

See #190 and #191

> As discussed, a user should be able to check their fingerprint. I believe the user profile would make most sense?

I guess this can be part of *this* PR.

SailReal:
User Profile
image

---

It would be nice to be able to see the full fingerprint in the Grant Access dialogue when hovering over it,  but this does not look like it would be easy to implement.

SailReal:
With the power of @JaniruTEC we now show the full fingerprint on hover:
image

JaniruTEC:
Since I am now officially part of this PR (😁), I have some minor ideas/suggestions:
- Add a small fingerprint icon (maybe from Font Awesome?) next to the fingerprints in the "Grant Permission" dialog
- Divide the full fingerprint shown on hover into groups similar to how it's done on the user profile page
- Find some place to spell out the word "User Fingerprint"

SailReal:
This is just a very basic way to verify user fingerprint and will completely replaced by https://github.com/cryptomator/hub/issues/191 in the next minor version so I would invest the time to enhance it into the replacing implementation.

> Add a small fingerprint icon (maybe from Font Awesome?) next to the fingerprints in the "Grant Permission" dialog

I decided against it because there is not much place in this dialog especially because usernames often contains more characters like test1 if we have e.g. julian.raufelder@foo.bar.baz there isn't that much space left in this dialog.

> Divide the full fingerprint shown on hover into groups similar to how it's done on the user profile page

This would require another variable that I wanted to avoid :sweat_smile: 

> Find some place to spell out the words "User Fingerprint"

Where would you expect those words in this dialog?

JaniruTEC:
> Where would you expect those words in this dialog?

I'd expect them probably on hover, either over the aforementioned icon or over the short fingerprint. Something like:

`4242424242` with the hover:
> baz' user fingerprint:
> 42424242424242424242424242424242424242

IMO many users will have no idea what those cryptic numbers mean -- unless they make the connection to their own user fingerprint, which is something I wouldn't count on.

infeo (approved):

infeo:
Because of #190 and #191 we will rework this anyway. For now it is important, that the user is able to view and check the fingerprints in a way.

> IMO many users will have no idea what those cryptic numbers mean -- unless they make the connection to their own user fingerprint, which is something I wouldn't count on.

We should mention it in the docs at least.

PR #238, opened 2023-11-07
Rename SetupCode to AccountKey in translation keys

SailReal:
As discussed, user-facing stuff, including translation keys, is now called "Account Key". Everywhere else in the code it is called "Setup Code".


frontend/src/components/InitialSetup.vue
```diff
@@ -88,7 +88,7 @@
                   <input id="confirmSetupKey" v-model="confirmSetupKey" name="confirmSetupKey" type="checkbox" class="h-4 w-4 rounded border-gray-300 text-primary focus:ring-primary" required>
```
tobihagemann, line 88: I noticed a "Setup Key" here. :smile:

SailReal: Good catch. Impressive how many variations of this term we had.

overheadhunter (changes requested):
Isn't it confusing if the translation *keys* don't match what they technically belong to?

I'd rather keep it as it (well rename setupKey to setupCode is fine of course)

tobihagemann:
It's confusing either way. I prefer not to confuse translators because they actually use the key as context. We have already received reports that the key looks wrong.

But since it's confusing either way, I'm also fine with the other option.

overheadhunter (approved):

PR #239, opened 2023-11-08
Sync Cryptomator Hub CLI client user to Hub

SailReal:
For Cryptomator Hub CLI to work, we need to have the user account of the client available in Hub.

If the Cryptomator Hub CLI is used, you need to

1. create a client with client id `cryptomatorhub-cli` and set `Service accounts roles`
2. add the `view-clients` role to the `syncer` realm role

### DIFF
--- a/backend/src/main/java/org/cryptomator/hub/KeycloakRemoteUserProvider.java
+++ b/backend/src/main/java/org/cryptomator/hub/KeycloakRemoteUserProvider.java
@@ -44,7 +44,22 @@ public class KeycloakRemoteUserProvider implements RemoteUserProvider {
 			users.addAll(currentRequestedUsers);
 		} while (currentRequestedUsers.size() == MAX_COUNT_PER_REQUEST);
 
+		var cliUser = cryptomatorCliUser(realm);
+		cliUser.ifPresent(users::add);
+
 		return users;
+	}
+
+	//visible for testing
+	Optional<User> cryptomatorCliUser(RealmResource realm) {
+		var clients = realm.clients().findByClientId("cryptomatorhub-cli");
+		if (clients.isEmpty()) {
+			return Optional.empty();
+		}
+		var clientId = clients.get(0).getId();
+		var client = realm.clients().get(clientId);
+		var clientUser = client.getServiceAccountUser();
+		return Optional.of(mapToUser(clientUser));
 	}
 
 	private Predicate<UserRepresentation> notSyncerUser() {
--- a/backend/src/main/resources/dev-realm.json
+++ b/backend/src/main/resources/dev-realm.json
@@ -177,6 +177,7 @@
       "clientId": "cryptomatorhub-cli",
       "name": "Cryptomator Hub CLI",
       "clientAuthenticatorType": "client-secret",
+      "secret": "top-secret",
       "standardFlowEnabled": false,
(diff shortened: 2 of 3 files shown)


overheadhunter (changes requested):
Works as intended, but shouldn't we also create this client via the dev realm json?

Maybe we can even set a fixed uuid for the service account user for easier testing.

overheadhunter (approved):

overheadhunter:
Sollten dann den realm-json-Generator auch aktualisieren (würde auch in älteren Hub-Versionen nicht stören)

SailReal:
> but shouldn't we also create this client via the dev realm json?

Totally agree, but should we also set a fixed secret so that we can reuse our CLI calls and not have to look up the secret in the dev realm like with the admin password? Something like `"secret": "SmbrC4S4O4qVtOA0uLJtrUHuAbJpvzFZ"`

> Sollten dann den realm-json-Generator auch aktualisieren (würde auch in älteren Hub-Versionen nicht stören)

But IMO the default shouldn't be touched because I assume 99.9% won't use the CLI tool and for the others we could introduce a checkbox that if checked will add those changes to the realm-json.

overheadhunter:
> > but shouldn't we also create this client via the dev realm json?
> 
> 
> 
> Totally agree, but should we also set a fixed secret in the dev realm so that we can reuse our CLI calls and not have to look up the secret like with the admin password? Something like `"secret": "top-secret"`

Agree

> 
> 
> 
> > Sollten dann den realm-json-Generator auch aktualisieren (würde auch in älteren Hub-Versionen nicht stören)
> 
> 
> 
> But IMO the default shouldn't be touched because I assume 99.9% won't use the CLI tool and for the others we could introduce a checkbox that if checked will add those changes to the realm-json.

Disagree 😉 

Without a fixed secret of course, then the client is there if needed but wont hurt. Maybe it can be added and be disabled?

SailReal:
> > > Sollten dann den realm-json-Generator auch aktualisieren (würde auch in älteren Hub-Versionen nicht stören)
> >
> >
> >
> > But IMO the default shouldn't be touched because I assume 99.9% won't use the CLI tool and for the others we could introduce a checkbox that if checked will add those changes to the realm-json.
> 
> Disagree 😉 
> 
> Without a fixed secret of course, then the client is there if needed but wont hurt. Maybe it can be added and be disabled?

A deactivated client would not do any harm, that is correct, but in order for the syncer to have access to the clients, its permissions must be extended (`view-clients`) and I would want to avoid that. If we do not set the permission, support gets too much complicated IMO, you then have the following states:
1. Old instances where client is not set and syncer has no `view-clients` permission
2. New instances that have disabled client with or without `view-clients` permission
3.  New instances that have enabled client with `view-clients` permission
3.  New instances that have enabled client without `view-clients` permission

I'll test this, but it would be cool to offer a partial realm import where the client is created and the syncer gets more permissions if a user wants to use Cryptomator Hub CLI. Existing clients could then use this as well. See if it works.

overheadhunter:
Ok let's leave it as is, adding the client manually if needed (maybe add an article to the docs).

Nevertheless I guess we should eventually replace the syncer user with a system client, which has further permissions, allowing stuff like managing certain Keycloak entities directly via Hub's UI.

PR #242, opened 2023-11-13
refactored `PUT /api/vaults/{v}/access-tokens/{u}` → `POST /api/vaults/{v}/access-tokens`

overheadhunter:
The new API accepts multiple access grants in a single requests, making bulk requests easier, as can be seen here:

https://github.com/cryptomator/hub/blob/1c2133d9868a2de3922aaa73b7b1cf3d90063cca/frontend/src/components/GrantPermissionDialog.vue#L127-L135


SailReal (approved):
LGTM; As discussed, we should add in the future an HTTP status code 207 like in WebDAV but we can enhance that later.

PR #245, opened 2023-11-24
Redesigned unlock success

overheadhunter:
Replaced the green "welcome back, alice" screens with more useful messages and matching the design of the remaining page.

Also added a new state when unlocking a vault after migrating to 1.3.0, when the account requires setting up user keys.

> [!note]
> Not yet localized, as we require to guess the browser language, first.

## when oauth failed
Bildschirmfoto 2023-11-24 um 13 38 46

## when user didn't complete setup yet
Bildschirmfoto 2023-11-24 um 13 38 56

## when the app needs registration
either on a known browser:
Bildschirmfoto 2023-11-24 um 13 37 53

or when the browser is unknown:
Bildschirmfoto 2023-11-24 um 13 37 56

## 403
Bildschirmfoto 2023-11-24 um 13 41 08

## success
Bildschirmfoto 2023-11-24 um 13 40 09

### DIFF
--- a/frontend/src/components/UnlockError.vue
+++ b/frontend/src/components/UnlockError.vue
@@ -1,22 +1,24 @@
 <template>
-  <div class="max-w-7xl mx-auto sm:px-6 lg:px-8">
-    <div class="pt-8 pb-4 shrink-0 flex items-center">
-      <img src="/logo.svg" class="h-8" alt="Logo"/>
-      <span class="font-headline font-bold text-primary ml-2 pb-px">CRYPTOMATOR HUB</span>
+  <nav class="bg-tertiary2">
+    <div class="max-w-7xl mx-auto sm:px-6 lg:px-8 h-16 flex-1 flex items-center justify-center sm:items-stretch sm:justify-start">
+      <div class="shrink-0 flex items-center">
+        <img src="/logo.svg" class="h-8" alt="Logo"/>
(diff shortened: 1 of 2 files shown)


SailReal (approved):
LGTM :+1:

Just one thing I noticed and wanted to mention, but I'm totally fine with, is that with a successful legacy device unlock, the "when user didn't complete setup yet" variant is shown, but this is perfectly fine imo, because even if the user has set up their user keys, they can still use the legacy unlock.

infeo:
I suggest to change the wording of the "Setup Required" screen, the "few simple steps" sounds very vague and not reassuring. I plead for keeping it simple:
> To unlock, you need to setup your account.
And then also update the button description, because "Complete setup" sounds like the user already started:
> Setup Account

overheadhunter:
> I suggest to change the wording of the "Setup Required" screen, the "few simple steps" sounds very vague and not reassuring. I plead for keeping it simple:
> 
> > To unlock, you need to setup your account.

Actually, legacy unlock may work, as Julian mentioned above. Nevertheless users shall complete setting up the account keys.

But I'll get rid of the "few simple steps", it really is just a single step anyway.

> 
> And then also update the button description, because "Complete setup" sounds like the user already started:
> 
> > Setup Account

This also affects users who already have an account, thus it would sound as if their account was deleted. Thus I'll keep the "Complete" metaphor to not slow this PR down, but am open for other suggestions.

Issue #109, opened 2022-08-09
Request new license token if old one is expired.


cryptobot:
If the license token expires, a new token should be accquired with the {{refreshUrl}} claim in the old token.
Two strategies are possible:

* automatic: Use polling or a predefined trigger to get new token
* manual: Show a refresh button in the billing component

Of course, the first is the desirable way, increasing UX by not forcing the user to act and keeping it convenient.

Remark: The {{refreshUrl}} must be called in backend due to CORS restrictions in frontend.

tobihagemann:
I'd like to add that I changed the refreshUrl in the store. Older tokens still point to `…/subscription` but new ones point to `…/token`. Redirection is in place, i.e., redirects have to be followed when implementing this. :wink:

PR #246, opened 2023-11-27
Refresh hub license daily at 01:00 UTC

SailReal:
Related to https://github.com/cryptomator/hub/issues/109

### DIFF
--- a/backend/src/main/java/org/cryptomator/hub/license/LicenseHolder.java
+++ b/backend/src/main/java/org/cryptomator/hub/license/LicenseHolder.java
@@ -3,6 +3,7 @@ package org.cryptomator.hub.license;
 import com.auth0.jwt.exceptions.JWTVerificationException;
 import com.auth0.jwt.interfaces.Claim;
 import com.auth0.jwt.interfaces.DecodedJWT;
+import io.quarkus.scheduler.Scheduled;
 import jakarta.annotation.PostConstruct;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
@@ -11,9 +12,18 @@ import org.cryptomator.hub.entities.Settings;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.jboss.logging.Logger;
 
+import java.io.IOException;
+import java.net.URI;
+import java.net.URLEncoder;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.nio.charset.StandardCharsets;
 import java.time.Instant;
+import java.util.HashMap;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 @ApplicationScoped
 public class LicenseHolder {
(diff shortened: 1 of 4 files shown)


backend/src/main/java/org/cryptomator/hub/license/LicenseHolder.java
```diff
@@ -75,6 +85,45 @@ public void set(String token) throws JWTVerificationException {
...
+				var client = HttpClient.newBuilder().followRedirects(HttpClient.Redirect.NORMAL).build();
+				refreshLicense(refreshUrl.get(), license.getToken(), client);
+			}
+		}
+	}
+
+	//visible for testing
+	void refreshLicense(String refreshUrl, String license, HttpClient client) throws InterruptedException {
+		var parameters = new HashMap<String, String>();
+		parameters.put("token", license);
```
overheadhunter, line 105: Use `Map.of(...)`

backend/src/main/java/org/cryptomator/hub/license/LicenseHolder.java
```diff
@@ -75,6 +85,45 @@ public void set(String token) throws JWTVerificationException {
...
+		var body = parameters.entrySet() //
+				.stream() //
+				.map(e -> e.getKey() + "=" + URLEncoder.encode(e.getValue(), StandardCharsets.UTF_8)) //
```
overheadhunter, line 108: I am not sure if URLEncoder does the right thing here. If I am not mistaken, it encodes stuff like ` ` to `%20` but it *does not* encode characters that are allowed in URLs, such as `&` and `=`.

What we really need is an encoder that encodes every character which is not allowed in *query parameters*.

Add a unit test which tests this method with `license = "foo&bar=baz"`.

backend/src/main/java/org/cryptomator/hub/license/LicenseHolder.java
```diff
@@ -75,6 +85,45 @@ public void set(String token) throws JWTVerificationException {
...
+				.collect(Collectors.joining("&"));
+		var request = HttpRequest.newBuilder() //
+				.uri(URI.create(refreshUrl)) //
+				.headers("Content-Type", "application/x-www-form-urlencoded") //
+				.POST(HttpRequest.BodyPublishers.ofString(body))  //
```
overheadhunter, line 113: Isn't there a body publisher for form-urlencode that handles all the encoding for us?

overheadhunter (changes requested):

backend/src/main/java/org/cryptomator/hub/license/LicenseHolder.java
SailReal: AFAIK it does exactly what we want, from https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/net/URLEncoder.html

> When encoding a String, the following rules apply:
> * The alphanumeric characters "a" through "z", "A" through "Z" and "0" through "9" remain the same.
> * The special characters ".", "-", "*", and "_" remain the same.
> * The space character "   " is converted into a plus sign "+".
> * All other characters are unsafe and are first converted into one or more bytes using some encoding scheme. > Then each byte is represented by the 3-character string "%xy", where xy is the two-digit hexadecimal representation of the byte. The recommended encoding scheme to use is UTF-8. However, for compatibility reasons, if an encoding is not specified, then the default encoding of the platform is used. 
> 
> For example using UTF-8 as the encoding scheme the string "The string ü@foo-bar" would get converted to "The+string+%C3%BC%40foo-bar" because in UTF-8 the character ü is encoded as two bytes C3 (hex) and BC (hex), and the character @ is encoded as one byte 40 (hex).

Added a unit test in [174c7aa](https://github.com/cryptomator/hub/pull/246/commits/174c7aad99b66f5373fc5c46b696b96ec41ad1d2)

backend/src/main/java/org/cryptomator/hub/license/LicenseHolder.java
SailReal: Fixed in [174c7aa](https://github.com/cryptomator/hub/pull/246/commits/174c7aad99b66f5373fc5c46b696b96ec41ad1d2)

backend/src/main/java/org/cryptomator/hub/license/LicenseHolder.java
SailReal: I researched for a while but couldn't find one...

overheadhunter: Apparently, there isn't. This [isn't the first time](https://github.com/coffeelibs/tiny-oauth2-client/blob/1bcba9b4e755fd7b97585c3375e2074994f5e864/src/main/java/io/github/coffeelibs/tinyoauth2client/TinyOAuth2Client.java#L183) I looked it up, I feel like I should add a feature request to the JDK.

backend/src/main/java/org/cryptomator/hub/license/LicenseHolder.java
overheadhunter: Nevermind, URLEncoder seems perfectly fitting for this job. Such a unit test would be a good addition anyway :)

SailReal: There is one, I changed the first test scenario to verify it: https://github.com/cryptomator/hub/blob/refresh-license/backend/src/test/java/org/cryptomator/hub/license/LicenseHolderTest.java#L221-L256

There I'm requesting a license for `token&foo=bar` which results in the encoded request body `token=token%26foo%3Dbar`.

overheadhunter (approved):

PR #247, opened 2023-11-29
Fix license handling

SailReal:


backend/src/main/java/org/cryptomator/hub/license/RandomMinuteSleeper.java
```diff
@@ -0,0 +1,17 @@
+package org.cryptomator.hub.license;
+
+import jakarta.enterprise.context.ApplicationScoped;
+
+import java.util.Random;
+
+@ApplicationScoped
+public class RandomMinuteSleeper {
```
tobihagemann, line 8: I wonder if there is a better name for it because I wouldn't understand that it would sleep between 0 and 60 minutes. But this shouldn't be a blocker, just wanted to mention it. We don't want to delay the release because of this.

tobihagemann (changes requested):

backend/src/main/java/org/cryptomator/hub/license/LicenseHolder.java
```diff
@@ -83,15 +100,16 @@ public void set(String token) throws JWTVerificationException {
 		var settings = Settings.get();
 		this.license = licenseValidator.validate(token, settings.hubId);
 		settings.licenseKey = token;
-		settings.persist();
+		settings.persistAndFlush();
 	}
 
 	/**
-	 * Attempts to refresh the Hub licence every day at 01:00 UTC if claim refreshURL is present.
+	 * Attempts to refresh the Hub licence every day at 01:00:00 UTC if claim refreshURL is present.
```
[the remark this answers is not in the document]
SailReal: Fixed in [913d636](https://github.com/cryptomator/hub/pull/247/commits/913d636bf40d03cd39a4c03a960fab5c9cd11518)

backend/src/main/java/org/cryptomator/hub/license/RandomMinuteSleeper.java
SailReal: I don't think the name is perfect either, but I couldn't find a better one yesterday and just now. I'm open to suggestions :)

tobihagemann: Maybe something like `RandomIntervalSleeper`? But then you would have to add a param to `sleep()` like `maxMinutesInterval`. Not sure if this is better.

tobihagemann (approved):

tobihagemann (approved):

Issue #222, opened 2023-08-17
Revert native sql unlock before we ship the refactored access management


SailReal:
### Please agree to the following

- [X] I have searched [existing issues](https://github.com/cryptomator/hub/issues?q=) for duplicates
- [X] I agree to follow this project's [Code of Conduct](https://github.com/cryptomator/hub/blob/develop/.github/CODE_OF_CONDUCT.md)

### Summary

Revert https://github.com/cryptomator/hub/commit/1baf7340870093676e7476331b6e94e417917518 before we ship the refactored access management to remove the native call which can introduce major problems in the future.

### Motivation

Make sure that the introduced native unlock call https://github.com/cryptomator/hub/commit/1baf7340870093676e7476331b6e94e417917518 never makes it into production. I strongly suspect that Hibernate ORM will be at version 6.2.8 or 6.3.0 until we ship the refactored access management.

This issue should just remind us to undo that commit before shipping.

### Considered Alternatives

We should not downgrade Quarkus to v3.1.0 because of https://github.com/cryptomator/hub/issues/220

Downgrading only the hibernate dependency wasn't successfully. 

### Anything else?

Regarding the issue itself, see https://github.com/quarkusio/quarkus/issues/35386

SailReal:
Quarkus 3.4.1 fixes it: https://github.com/quarkusio/quarkus/blob/6c6c3b972b34268a34d5ca8f55858d494742ac0e/bom/application/pom.xml#L102

PR #252, opened 2024-01-25
Add new API for Legacy Device Migration

overheadhunter:
This adds two new API endpoints:

* deprecated `GET /api/devices/{deviceId}/legacy-access-tokens` (to be removed after all users migrated from 1.2.x)
* permanent `POST /api/users/me/access-tokens`

The response body of the former and the request body of the latter are both JSON objects with these key-value-pairs:

```json
{
  "vault1": "token1",
  "vault2": "token2",
  ...
}
```

The difference is that the deprecated legacy access tokens are encrypted using the device key, while the current access tokens posted to the users resource shall be encrypted using the user key.

Both endpoints are accessible to all users, as they only expose information owned by the currently logged in user.

## Summary by CodeRabbit

- **New Features**
	- Added the ability to list legacy access tokens for devices.
	- Introduced a new feature to update user-specific vault keys.

- **Bug Fixes**
	- Improved the retrieval of legacy access tokens by ensuring they are filtered by device and owner.

- **Tests**
	- Implemented new tests to verify the functionality of legacy access token retrieval.
	- Added tests to ensure proper handling of access token POST requests.

### DIFF
--- a/backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
+++ b/backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
@@ -6,6 +6,7 @@ import jakarta.inject.Inject;
 import jakarta.transaction.Transactional;
 import jakarta.validation.Valid;
 import jakarta.validation.constraints.NotEmpty;
+import jakarta.validation.constraints.NotNull;
 import jakarta.ws.rs.Consumes;
 import jakarta.ws.rs.GET;
 import jakarta.ws.rs.POST;
@@ -74,7 +75,7 @@ public class UsersResource {
 	@Consumes(MediaType.APPLICATION_JSON)
 	@Operation(summary = "adds/updates user-specific vault keys", description = "Stores one or more vaultid-vaultkey-tuples for the currently logged-in user, as defined in the request body ({vault1: token1, vault2: token2, ...}).")
 	@APIResponse(responseCode = "200", description = "all keys stored")
-	public Response updateMyAccessTokens(@NotEmpty Map<UUID, String> tokens) {
+	public Response updateMyAccessTokens(@NotNull Map<UUID, String> tokens) {
 		var user = User.<User>findById(jwt.getSubject());
 		for (var entry : tokens.entrySet()) {
 			var vault = Vault.<Vault>findById(entry.getKey());
--- a/backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
+++ b/backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
@@ -30,6 +30,7 @@ import java.time.temporal.ChronoUnit;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.UUID;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
@@ -73,7 +74,7 @@ public class UsersResource {
 	@Consumes(MediaType.APPLICATION_JSON)
 	@Operation(summary = "adds/updates user-specific vault keys", description = "Stores one or more vaultid-vaultkey-tuples for the currently logged-in user, as defined in the request body ({vault1: token1, vault2: token2, ...}).")
 	@APIResponse(responseCode = "200", description = "all keys stored")
-	public Response updateMyAccessTokens(@NotEmpty Map<String, String> tokens) {
+	public Response updateMyAccessTokens(@NotEmpty Map<UUID, String> tokens) {
 		var user = User.<User>findById(jwt.getSubject());
 		for (var entry : tokens.entrySet()) {
 			var vault = Vault.<Vault>findById(entry.getKey());
--- a/backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
+++ b/backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
@@ -78,7 +78,7 @@ public class UsersResource {
 		var user = User.<User>findById(jwt.getSubject());
 		for (var entry : tokens.entrySet()) {
 			var vault = Vault.<Vault>findById(entry.getKey());
-			if (vault == null || vault.archived) {
+			if (vault == null) {
 				continue; // skip
 			}
 			var token = AccessToken.<AccessToken>findById(new AccessToken.AccessId(user.id, vault.id));
--- a/backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
+++ b/backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
@@ -5,6 +5,7 @@ import jakarta.annotation.security.RolesAllowed;
 import jakarta.inject.Inject;
 import jakarta.transaction.Transactional;
 import jakarta.validation.Valid;
+import jakarta.validation.constraints.NotEmpty;
 import jakarta.ws.rs.Consumes;
 import jakarta.ws.rs.GET;
 import jakarta.ws.rs.POST;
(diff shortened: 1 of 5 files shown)


backend/src/main/java/org/cryptomator/hub/entities/LegacyAccessToken.java
```diff
@@ -22,6 +23,13 @@
 		INNER JOIN effective_vault_access a ON a.vault_id = t.vault_id AND a.authority_id = d.owner_id
 		WHERE t.vault_id = :vaultId AND d.id = :deviceId AND d.owner_id = :userId
 		""")
+@NamedNativeQuery(name = "LegacyAccessToken.getByDevice", resultClass = LegacyAccessToken.class, query = """
```
infeo, line 26: Why are we using a NativeQuery here?

infeo: I see that we are already using it for the other method, but the regarding commit (https://github.com/cryptomator/hub/commit/491739a7c3b372886f23b4a9a1608f0d2ca94d9c) does not give any insight about the change reason.

backend/src/test/java/org/cryptomator/hub/api/DeviceResourceTest.java
infeo, line 1: Can we have also a `GET /devices/legacyDevice2/legacy-access-tokens  returns 404` here? Or does the API then return an empty list?

backend/src/test/java/org/cryptomator/hub/api/UsersResourceTest.java
```diff
@@ -70,6 +69,20 @@ public void testGetAll() {
 					.body("id", hasItems("user1", "user2"));
 		}
 
+		@Test
+		@DisplayName("POST /users/me/access-tokens returns 200")
+		public void testPostAccessTokens() {
+			var body = """
+					{
+						"7E57C0DE-0000-4000-8000-000100001111": "jwe.jwe.jwe.vault1.user1",
+						"7E57C0DE-0000-4000-8000-BADBADBADBAD": "noSuchVault"
```
infeo, line 78: Add a vault, which exists and is not owned by the test user.

infeo (changes requested):

backend/src/main/java/org/cryptomator/hub/entities/LegacyAccessToken.java
overheadhunter: Good observation. @SailReal do you recall why we chose to use a native query? I know it was a workaround for some weird bug, but git blame is not very helpful in particular.

backend/src/test/java/org/cryptomator/hub/api/DeviceResourceTest.java
overheadhunter: it should return an empty list, but I can add a test anyway

backend/src/test/java/org/cryptomator/hub/api/UsersResourceTest.java
overheadhunter: Do you mean _not accessible_?

backend/src/main/java/org/cryptomator/hub/entities/LegacyAccessToken.java
SailReal: It is because of https://github.com/cryptomator/hub/issues/222

overheadhunter: I see, but the questionable joined inheritance is no longer part of these queries, so I guess the issue might be obsolete? I'll check the generated native queries.

overheadhunter: I did some promising experiments but I believe solving #222 should be a separate PR.

backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
```diff
@@ -62,6 +67,33 @@ public Response putMe(@Nullable @Valid UserDto dto) {
...
+	@RolesAllowed("user")
+	@Transactional
+	@Consumes(MediaType.APPLICATION_JSON)
+	@Operation(summary = "adds/updates user-specific vault keys", description = "Stores one or more vaultid-vaultkey-tuples for the currently logged-in user, as defined in the request body ({vault1: token1, vault2: token2, ...}).")
+	@APIResponse(responseCode = "200", description = "all keys stored")
+	public Response updateMyAccessTokens(@NotEmpty Map<UUID, String> tokens) {
+		var user = User.<User>findById(jwt.getSubject());
+		for (var entry : tokens.entrySet()) {
+			var vault = Vault.<Vault>findById(entry.getKey());
+			if (vault == null || vault.archived) {
```
SailReal, line 81: I'm not quite sure if we want to skip archived vaults during migration. The reason is probably that you don't want to touch it because it's archived, but as far as unlocking is concerned, the migration doesn't change anything, it needs to be migrated later when the vault is unarchived again. I think our goal should be to have as few legacy tokens in the system as possible, or what do you think?

SailReal (commented):
LGTM; just one question :)

backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
overheadhunter: I wasn't sure either. Lets decide this in a short call.

overheadhunter: Decision: We will allow updating access tokens for _every_ accessible vault, regardless of whether it is archived or not. Unlock will of course still be denied for archived vaults, but data migration is in the spirit of the decision that led to archiving (as opposed to deleting): We want to avoid loss of data.

backend/src/test/java/org/cryptomator/hub/api/UsersResourceTest.java
overheadhunter: Technically there is nothing wrong if you add access tokens for yourself, even if lack permissions to access the vault.

Since @SailReal and I decided to make this part of the API a permanent one that might be useful in other cases in which a user might want to prepare key ownership before being granted access, I plan to keep it this way.

backend/src/test/java/org/cryptomator/hub/api/DeviceResourceTest.java
overheadhunter: added test in ff0967edb07419bd34e5dd088e620ac2f950143a

backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
```diff
@@ -62,6 +67,33 @@ public Response putMe(@Nullable @Valid UserDto dto) {
...
 	}
 
+	@POST
+	@Path("/me/access-tokens")
+	@RolesAllowed("user")
+	@Transactional
+	@Consumes(MediaType.APPLICATION_JSON)
+	@Operation(summary = "adds/updates user-specific vault keys", description = "Stores one or more vaultid-vaultkey-tuples for the currently logged-in user, as defined in the request body ({vault1: token1, vault2: token2, ...}).")
+	@APIResponse(responseCode = "200", description = "all keys stored")
+	public Response updateMyAccessTokens(@NotEmpty Map<UUID, String> tokens) {
```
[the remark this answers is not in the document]
infeo: It is important, that if the user does not have any access tokens, he/she does not get an error returned. Otherwise our client application has to deal with this edge case.

infeo (changes requested):

backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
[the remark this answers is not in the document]
overheadhunter: > nor do i see the necessity to be non-empty

Indeed it will be a no-op. I'll remove this constraint (copy-pasted anyway)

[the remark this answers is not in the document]
overheadhunter: changed in 58ff0d684ba1eda836320430e58b44b48029baff

infeo (approved):

PR #264, opened 2024-03-15
Feature: License status endpoint

### LINKED ISSUE #262: On License exceedance vault creation possible

### Summary

If the license is exceeded, it should not be possible to create a vault. Still the create vault flow can be started, but stops with an error. If the user returns to the vault list, the vault is created anyway.s

### System Setup

```markdown
- Hub: 1.3.3
- Keycloak: -
```

### Steps to Reproduce

1. Ensure that license is exceeded (e.g. add group to a vault and increase group size afterwards with new members)
2. log in as a user or admin
3. Start create vault flow

### Expected Behavior

Cannot start vault creation, because license is exceeded.

### Actual Behavior

Can continue vault creation till display of recovery key:
grafik

Returing to the vault list, the vault unexpectedly exists:
grafik

### Reproducibility

Always

### Relevant Log Output

_No response_

### Anything else?

_No response_

infeo:
Fixes #262.

This PR adds the new REST endpoint `/license`.

This endpoint is intended to give information about the license set in the hub instance. Currently, this endpoint does not have any direct actions, but only a subresource `/license/user-info`. This is intentional, because the longterm goal is to move the `/billing` endpoint to `/license`.

The subresource `/license/user-info` is intended for any user to get necessary information about the license like expiration date and licensed seats.

Remark: The license endpoint could also be used in the `VaultDetails` vue component, but due to a required refactoring there it is not used.

## Summary by CodeRabbit

- **New Features**
    - Introduced a new `LicenseResource` class to provide endpoints for retrieving license status information.
    - Added `LicenseStatusDto` and `LicenseService` in the frontend to support license status checks and display.
    - Enhanced the Vault List view to show `LicenseAlert` based on license status.

- **Bug Fixes**
    - Corrected the license seats count logic in the billing system.

- **Refactor**
    - Improved logging in `LicenseHolder`.
    - Renamed and updated methods for clarity and testing compatibility.

- **Style**
    - Updated `MenuButton` attributes in `VaultList.vue` for better user interaction based on license status.

### DIFF
--- a/backend/src/main/java/org/cryptomator/hub/api/LicenseResource.java
+++ b/backend/src/main/java/org/cryptomator/hub/api/LicenseResource.java
@@ -26,7 +26,7 @@ public class LicenseResource {
 	@Path("/status")
 	@Produces(MediaType.APPLICATION_JSON)
 	@RolesAllowed("user")
-	@Operation(summary = "Get license status information", description = "Information includes the license expiration date, the licensed seats and the already used seats")
+	@Operation(summary = "Get license status information", description = "Information includes the licensed seats, the already used seats and if defined, the license expiration date.")
 	@APIResponse(responseCode = "200")
 	public LicenseStatusDto get() {
 		return LicenseStatusDto.create(licenseHolder);
--- a/backend/src/main/java/org/cryptomator/hub/api/LicenseResource.java
+++ b/backend/src/main/java/org/cryptomator/hub/api/LicenseResource.java
@@ -23,25 +23,25 @@ public class LicenseResource {
 	LicenseHolder licenseHolder;
 
 	@GET
-	@Path("/status")
+	@Path("/user-info")
 	@Produces(MediaType.APPLICATION_JSON)
 	@RolesAllowed("user")
-	@Operation(summary = "Get license status information", description = "Information includes the licensed seats, the already used seats and if defined, the license expiration date.")
+	@Operation(summary = "Get license information for regular users", description = "Information includes the licensed seats, the already used seats and if defined, the license expiration date.")
 	@APIResponse(responseCode = "200")
-	public LicenseStatusDto get() {
-		return LicenseStatusDto.create(licenseHolder);
+	public LicenseUserInfoDto get() {
+		return LicenseUserInfoDto.create(licenseHolder);
 	}
 
 
-	public record LicenseStatusDto(@JsonProperty("licensedSeats") Integer licensedSeats,
-								   @JsonProperty("usedSeats") Integer usedSeats,
-								   @JsonProperty("expiresAt") Instant expiresAt) {
+	public record LicenseUserInfoDto(@JsonProperty("licensedSeats") Integer licensedSeats,
+									 @JsonProperty("usedSeats") Integer usedSeats,
+									 @JsonProperty("expiresAt") Instant expiresAt) {
 
-		public static LicenseStatusDto create(LicenseHolder licenseHolder) {
+		public static LicenseUserInfoDto create(LicenseHolder licenseHolder) {
 			var licensedSeats = (int) licenseHolder.getSeats();
 			var usedSeats = (int) EffectiveVaultAccess.countSeatOccupyingUsers();
 			var expiresAt = Optional.ofNullable(licenseHolder.get()).map(DecodedJWT::getExpiresAtAsInstant).orElse(null);
-			return new LicenseStatusDto(licensedSeats, usedSeats, expiresAt);
+			return new LicenseUserInfoDto(licensedSeats, usedSeats, expiresAt);
 		}
 
 	}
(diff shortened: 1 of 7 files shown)


frontend/src/common/backend.ts
```diff
@@ -195,6 +195,23 @@ export type VersionDto = {
   keycloakVersion: string;
 }
 
+export class LicenseStatusDto {
+  constructor(
+    public licensedSeats: number,
+    public usedSeats: number,
+    public expiresAt: Date) {
```
SailReal, line 202: `expiresAt` might be `null`:

```suggestion
    public expiresAt?: Date) {
```

backend/src/main/java/org/cryptomator/hub/api/LicenseResource.java
```diff
@@ -0,0 +1,49 @@
...
+public class LicenseResource {
+
+	@Inject
+	LicenseHolder licenseHolder;
+
+	@GET
+	@Path("/status")
+	@Produces(MediaType.APPLICATION_JSON)
+	@RolesAllowed("user")
+	@Operation(summary = "Get license status information", description = "Information includes the license expiration date, the licensed seats and the already used seats")
```
SailReal, line 29:
```suggestion
	@Operation(summary = "Get license status information", description = "Information includes the license expiration date (optional), the licensed seats and the already used seats")
```

frontend/src/common/backend.ts
```diff
@@ -355,6 +372,14 @@ class BillingService {
   }
 }
 
+class LicenseService {
+  public async getStatus(): Promise<LicenseStatusDto> {
+    return axiosAuth.get('/license/status').then(response => {
+      return new LicenseStatusDto(response.data.licensedSeats, response.data.usedSeats, new Date(response.data.expiresAt));
```
SailReal, line 378: `new Date(null)` is the Unix Epoch timestamp but should be `null` imo.

```suggestion
      return new LicenseStatusDto(response.data.licensedSeats, response.data.usedSeats, response.data.expiresAt ? new Date(response.data.expiresAt) : null);
```

SailReal (changes requested):
LGTM :+1: , would only slightly improve the optional expiration date handling.

frontend/src/common/backend.ts
```diff
@@ -195,6 +195,23 @@ export type VersionDto = {
...
+export class LicenseStatusDto {
+  constructor(
+    public licensedSeats: number,
+    public usedSeats: number,
+    public expiresAt: Date) {
+  }
+
+  public isExpired(): boolean {
+    const now = new Date();
+    return now > this.expiresAt;
```
SailReal, line 207: Here we should also honor the optional expiration date.

SailReal:
Another thing I noticed: in https://github.com/cryptomator/hub/blob/feature/license-status-endpoint/frontend/src/components/LicenseAlert.vue we still use the billing resource, which means that still only admins see this alert, do we want to adjust this in a separate PR or in this one?

frontend/src/common/backend.ts
infeo: Fixed in f9c3f00f7f1d5197a054455571229cbe1ae3e716.

frontend/src/common/backend.ts
infeo: Fixed in f9c3f00f7f1d5197a054455571229cbe1ae3e716.

backend/src/main/java/org/cryptomator/hub/api/LicenseResource.java
infeo: Fixed in f9c3f00f7f1d5197a054455571229cbe1ae3e716.

frontend/src/common/backend.ts
infeo: Fixed in f9c3f00f7f1d5197a054455571229cbe1ae3e716.

infeo:
> means that still only admins see this alert, do we want to adjust this in a separate PR or in this one?

I changed it in fedd8de23e1a1f918e42a319c2c5dc5cfe814b5d

frontend/src/components/LicenseAlert.vue
```diff
@@ -15,7 +15,7 @@
     </div>
   </div>
 
-  <div v-else-if="admin && (admin.licensedSeats - admin.usedSeats < 0)" class="rounded-md bg-yellow-50 p-4 mb-3">
+  <div v-else-if="license && (license.licensedSeats - license.usedSeats < 0)" class="rounded-md bg-yellow-50 p-4 mb-3">
```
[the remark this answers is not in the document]
infeo: @coderabbitai You are wrong. The null/undefined check happens in the mentioned if expressions by first checking for the object itself and only afterwards checking for defined properties on the object

SailReal (approved):
LGTM :+1:

PR #271, opened 2024-04-19
Feature: Rework vaultDetails logic

infeo:
This PR reworks/fixes the logic in vault details.

Changes:
* vaultRole property is extended by a new value (`NONE`). An admin can access all vaults, but does not have the `MEMBER` role. This prevents errors done in future changes and keeps the model clean.
* vaultDetails component emits now license updates, such that the license alert is  displayed in the vault list, if on the server side the license changes
* aligned translations key names used in the vaultDetails component
* renamed RecoveryKeyDialog to DisplayRecoveryKeyDialog to be aligned with other dialog components
* reworked license conditions
    * use new license/user-info endpoint
    * if the license is violated (aka expired or exceeded), only block grantAccess and addMember
* seperated action bar into 4 containers
    * vault is a legaceVault and someone needs to claim ownership
    * vault is archived
    * vault recovery is required
    * other/regular vault
* if vault is archived, you can still edit _existing_ members (upgrade/downgrade/remove)
* make dialogs dependent only on required properties and the "show" variable


frontend/src/components/VaultDetails.vue
```diff
@@ -202,18 +213,19 @@ import FetchError from './FetchError.vue';
...
-import RecoveryKeyDialog from './RecoveryKeyDialog.vue';
+import DisplayRecoveryKeyDialog from './DisplayRecoveryKeyDialog.vue';
 import SearchInputGroup from './SearchInputGroup.vue';
 
 const { t, d } = useI18n({ useScope: 'global' });
 
 const props = defineProps<{
   vaultId: string,
-  role: VaultRole
+  vaultRole: string,
```
overheadhunter, line 223: Let's restrict this type to the set of allowed values:

```suggestion
  vaultRole: 'OWNER' | 'MEMBER' | 'NONE',
```

overheadhunter (changes requested):

frontend/src/components/VaultDetails.vue
```diff
@@ -202,18 +213,19 @@ import FetchError from './FetchError.vue';
...
-import RecoveryKeyDialog from './RecoveryKeyDialog.vue';
+import DisplayRecoveryKeyDialog from './DisplayRecoveryKeyDialog.vue';
 import SearchInputGroup from './SearchInputGroup.vue';
 
 const { t, d } = useI18n({ useScope: 'global' });
 
 const props = defineProps<{
   vaultId: string,
-  role: VaultRole
+  vaultRole: 'OWNER' | 'MEMBER' | 'NONE',
```
tobihagemann, line 223: Is there a reason, why you didn't keep the `VaultRole` type? It's defined in `backend.ts` and can also be extended by `'NONE'`.

frontend/src/i18n/en-US.json
```diff
@@ -220,13 +219,14 @@
   "vaultDetails.actions.updatePermissions.reload": "Reload",
   "vaultDetails.actions.editVaultMetadata": "Edit Vault Metadata",
   "vaultDetails.actions.downloadVaultTemplate": "Download Vault Template",
-  "vaultDetails.actions.showRecoveryKey": "Show Recovery Key",
+  "vaultDetails.actions.displayRecoveryKey": "Show Recovery Key",
   "vaultDetails.actions.archiveVault": "Archive Vault",
   "vaultDetails.actions.reactivateVault": "Reactivate Vault",
-  "vaultDetails.error.paymentRequired": "Your Cryptomator Hub license has expired or you exceeded the number of licensed seats. Please inform a Hub administrator renew or upgrade the license.",
+  "vaultDetails.actions.claimOwnership": "Become Vault Owner",
```
tobihagemann, line 225: Any reason why you changed this? I still prefer "Claim Ownership". It doesn't have to be a direct translation, I would even argue that English is the source.

frontend/src/components/VaultDetails.vue
```diff
@@ -108,90 +108,101 @@
...
-          </button>
-          <button type="button" class="bg-white py-2 px-4 border border-gray-300 rounded-md shadow-sm text-sm font-medium text-gray-700 hover:bg-gray-50 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-primary" @click="showRecoveryKeyDialog()">
-            {{ t('vaultDetails.actions.showRecoveryKey') }}
-          </button>
-          <button type="button" class="bg-red-600 py-2 px-4 border border-transparent rounded-md shadow-sm text-sm font-medium text-white  hover:bg-red-700 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-red-500" @click="showArchiveVaultDialog()">
-            {{ t('vaultDetails.actions.archiveVault') }}
-          </button>
-        </div>
+    <!-- button bar -->
+    <div >
```
tobihagemann, line 114: Not sure if this could be added as a linter rule, but `<div >` has a space. :smile:

tobihagemann (changes requested):
Just did a code review for now. Is my assumption right that you didn't really change anything in the UI? Or is there a noteworthy change?

frontend/src/components/VaultDetails.vue
infeo: Yes. 

In the backend the vaultRole `NONE` does not exist. The vaultRole field is only used in the `/vault/[vaultid]/members` endpoint and to keep the direct mapping in code, i did not touch `backend.ts`. Furthermore, in the context of the VaultList component, the role is determined indirect by checking the owned and accessible vaults and not using the VaultRole enum.

This keeps the model clean, but allows to differentiate in the GUI different cases.

frontend/src/components/VaultDetails.vue
infeo: Fixed in  79d980d37177daab6df0ff4cc75b24a85da2d41b

frontend/src/i18n/en-US.json
infeo: Reverted in 79d980d37177daab6df0ff4cc75b24a85da2d41b

infeo:
> Is my assumption right that you didn't really change anything in the UI? 

Nothing apparently visible. As wrote in the opening comment, i cleaned the license logic, such that only grant access and addMember are blocked. All other features are enabled.

SailReal (commented):
I really like this change!

Only I do not really understand the reason behind the change from `RecoveryKeyDialog` to `DisplayRecoveryKeyDialog`? For me a dialog is already something to display? Things get wild at the latest with  `showDisplayRecoveryKeyDialog` :sweat_smile:

infeo:
> Only I do not really understand the reason behind the change from RecoveryKeyDialog to DisplayRecoveryKeyDialog? 

If you are looking at the other `*Dialog` names, you notices a general scheme of "[Verb][Noun(s)]Dialog": A dialog does something with an object, aka it has an intention/action. A "RecoveryKeyDialog" breaks with this scheme, because it is not clear what the intention of the dialog is. SHould it change th recoveryKey? Should it display it?

I just fitted it into the scheme and from my perspective, it is now clearer and additionally easier to distinc from the `recoverVaultDialog`.

frontend/src/components/VaultDetails.vue
tobihagemann: Haven't tested it, but shouldn't be `VaultRole | 'NONE'` possible in that case?

infeo: It seems to work. Should VaultList.vue then also be adapted to use the vaultRole? Currently, there are only Strings used.
https://github.com/cryptomator/hub/blob/79d980d37177daab6df0ff4cc75b24a85da2d41b/frontend/src/components/VaultList.vue#L140-L148

tobihagemann: Yes, I've thought about this as well, but didn't comment on it. How would you approach this? Something like this?

```ts
const roleOfSelectedVault = computed<VaultRole | 'NONE'>(() => {
```

infeo: Done in https://github.com/cryptomator/hub/pull/271/commits/e4204c450f20c6797560e8dc53b0aa82340bcdc5

SailReal (changes requested):
In https://github.com/cryptomator/hub/blob/3a7a132dd8749b674bf031f51dc90daf5326ac78/frontend/src/components/VaultDetails.vue#L290

when the license has expired, the backend returns a 402 for `/vaults/${vaultId}/access-token?evenIfArchived=true` and is mapped to a `PaymentRequiredError` which is caught in this method, but results in no `vaultKeys` being set, and in turn the `Show Recovery Key` button cannot show the `DisplayRecoveryKeyDialog` but does nothing due to

https://github.com/cryptomator/hub/blob/3a7a132dd8749b674bf031f51dc90daf5326ac78/frontend/src/components/VaultDetails.vue#L191

We should not show this button if `vaultKeys` is missing.

image

infeo:
@SailReal Thanks for the finding. I fixed it in 2dd4d533f8ce5efcdadc239a3cc4bafb2c034ec4 creating an own `div` for actions possible, when the license is violated.

SailReal (approved):

tobihagemann (approved):

PR #281, opened 2024-05-24
Web of Trust

overheadhunter:
## API endpoints
This adds two new methods to the users resource:

1. `PUT /api/users/trusted/{userId}`, which stores a signature created by the current user for the user referenced by `userId`
2. `GET /api/users/trusted`, which retrieves all users with their corresponding signature chains that are directly or transitively trusted by the current user

[These added test cases](https://github.com/cryptomator/hub/blob/c71cfbe6ec643539dbabe8505a64fe87ccae0962/backend/src/test/java/org/cryptomator/hub/api/UsersResourceIT.java#L155-L237) demonstrate the usage.

## Database
This adds a new table `wot` as well as a view `effective_wot`. The `wot` table stores all signatures added via the aforementioned endpoint. The `effective_wot` contains the signature chains by looking up transitive trusts up to a graph depth of 10.

## Frontend
A component is added to signify whether a user can be trusted. It shows whether a chain of trust between the user in question and the logged in user exists. Furthermore it allows to sign the user's public keys.

## Future work
Currently, there are two new parameters in the `settings` table that may be exposed to the admin interface in a separate future PR:

1. `wot_max_depth` limits the number of edges in the WoT graph between two users when trying to find a trust chain
2. `wot_id_verify_len` is used by the frontend to specify how many hex digits of the hash of user U's public key a signer S has to enter in order to sign U's keys.

The chosen defaults are up for debate, of course.


overheadhunter:
@SailReal this is still WiP, but could you please review the backend part up to this point? This is merely the dumb storage logic, all signature creation and validation will then be done by the frontend.

The signature will simply be a JWT containing the trusted user's public key and is signed by the trusting user. Intermediate items in the signature chain can therefore be simply validated without further db lookups. Just the root and leaf public keys must be checked against their respective owners.

backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
```diff
@@ -149,4 +156,46 @@ public List<UserDto> getAll() {
...
+	public Response putSignature(@PathParam("userId") String userId, @NotNull String signature) {
+		var id = new WotEntry.Id();
+		id.setUserId(userId);
+		id.setSignerId(jwt.getSubject());
+		var entry = wotRepo.findById(id);
+		if (entry == null) {
+			entry = new WotEntry();
+			entry.setId(id);
+		}
+		entry.setSignature(signature);
```
SailReal, line 175: I don't think we should silently overwrite existing trusts but instead throw an 409 or something and then show to the user that there is already a trust entry and ask if the user wants to overwrite it. If the user passes a overwrite param to this endpoint it will be overwritten.

SailReal (changes requested):
LGTM, just the two mentioned topics.

backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
```diff
@@ -149,4 +156,46 @@ public List<UserDto> getAll() {
...
+		wotRepo.persist(entry);
```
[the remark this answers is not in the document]
overheadhunter: I agree but would probably create a separate PR after implementing the core functionality.

backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
overheadhunter: Imo, signing data is idempotent. It makes no difference if you look at an old or a new signature, except when you no longer trust the old signature algorithm. Silently overwriting allows updating a signature, if the need arises. Also, it reduces complexity as we don't need to add new dialogs for a 409 case.

backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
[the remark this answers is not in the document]
overheadhunter: added in f6ce0c0

backend/src/main/java/org/cryptomator/hub/entities/EffectiveWot.java
```diff
@@ -0,0 +1,108 @@
...
+		}
+
+		@Override
+		public String toString() {
+			return "EffectiveWotId{" +
+					"trustingUserId='" + trustingUserId + '\'' +
+					", trustedUserId='" + trustedUserId + '\'' +
+					'}';
+		}
+	}
```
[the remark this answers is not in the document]
overheadhunter: we don't use lombok in this project, please skip these suggestions in future reviews.

backend/src/main/java/org/cryptomator/hub/entities/StringArrayType.java
```diff
@@ -0,0 +1,66 @@
...
+	@Override
+	public Serializable disassemble(String[] value) {
+		return value; // value is immutable
+	}
+
+	@Override
+	public String[] assemble(Serializable cached, Object owner) {
+		return (String[]) cached; // value is immutable
+	}
+}
```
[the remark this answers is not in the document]
overheadhunter: Yes please. This class is indeed only used for the `effective_wot` view, i.e. in a read only manner.

backend/src/main/java/org/cryptomator/hub/api/SettingsResource.java
```diff
@@ -0,0 +1,62 @@
...
+	@Transactional
+	public Response put(@NotNull @Valid SettingsDto dto) {
+		var settings = settingsRepo.get();
+		settings.setWotMaxDepth(dto.wotMaxDepth);
+		settings.setWotIdVerifyLen(dto.wotIdVerifyLen);
+		settingsRepo.persist(settings);
+		return Response.status(Response.Status.NO_CONTENT).build();
+	}
+
+	public record SettingsDto(@JsonProperty("hubId") String hubId, @JsonProperty("wotMaxDepth") @Min(0) @Max(10) int wotMaxDepth, @JsonProperty("wotIdVerifyLen") @Min(0) int wotIdVerifyLen) {
```
SailReal, line 54: The current max of `wotMaxDepth` is 9, see https://github.com/cryptomator/hub/pull/281/files#diff-8237ab834f761b84d37b515e3bca1a1c40bf6821d87db8b39455f2c47bc59fbbR3

```suggestion
	public record SettingsDto(@JsonProperty("hubId") String hubId, @JsonProperty("wotMaxDepth") @Min(0) @Max(9) int wotMaxDepth, @JsonProperty("wotIdVerifyLen") @Min(0) int wotIdVerifyLen) {
```

backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
```diff
@@ -168,7 +177,65 @@ public Response resetMe() {
...
+
+	@GET
+	@Path("/trusted/{userId}")
+	@RolesAllowed("user")
+	@NoCache
+	@Transactional
+	@Produces(MediaType.APPLICATION_JSON)
+	@Operation(summary = "get trust detail for given user", description = "returns the shortest found signature chain for the given user")
+	@APIResponse(responseCode = "200")
+	@APIResponse(responseCode = "404")
```
SailReal, line 214: Should we add a description here?

backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
```diff
@@ -168,7 +177,65 @@ public Response resetMe() {
...
 	}
 
+	@GET
+	@Path("/trusted")
+	@RolesAllowed("user")
+	@NoCache
+	@Transactional
+	@Produces(MediaType.APPLICATION_JSON)
+	@Operation(summary = "get trusted users", description = "returns a list of users trusted by the currently logged-in user")
+	@APIResponse(responseCode = "200")
```
SailReal, line 227: Should we add a description here?

frontend/src/common/wot.ts
```diff
@@ -0,0 +1,101 @@
...
+    alg: 'ES384',
+    typ: 'JWT',
+    b64: true,
+    iss: me.id,
+    sub: user.id,
+    iat: Math.floor(Date.now() / 1000)
+  }, toSign, userKeys.ecdsaKeyPair.privateKey);
+  await backend.trust.trustUser(user.id, signature);
+  const trust = await backend.trust.get(user.id);
+  return trust!;
```
SailReal, line 42: Why this unwrap here? It can be actual undefined due to

https://github.com/cryptomator/hub/blob/99a527c2a1522e3b9cba71eb4886c197d732913c/frontend/src/common/backend.ts#L266C9-L266C57

frontend/src/components/AuditLogDetailsSignedWotId.vue
```diff
@@ -0,0 +1,57 @@
...
+
+const { t } = useI18n({ useScope: 'global' });
+
+const props = defineProps<{
+  event: AuditEventSignedWotIdDto
+}>();
+
+const resolvedUser = ref<UserDto>();
+const resolvedSigner = ref<UserDto>();
+const jwtLink = ref<string>(`https://jwt.io/#debugger-io?token=${props.event.signature}&publicKey=-----BEGIN%20PUBLIC%20KEY-----%0A${encodeURIComponent(props.event.signerKey)}%0A-----END%20PUBLIC%20KEY-----`);
```
SailReal, line 51: I think we should not use jwt.io here, it could raise privacy concerns, even though all the data here could be considered public. Could we not decode it inline and show it to the admin user directly in hub?

SailReal (changes requested):
Some things I came across:

1. The following is reproducible: Create trust chain admin-->alice, alice-->bob, bob-->carol, which works as expected until you deleted alice via Keyloak which breaks the syncer so I can not really verify if the chain is in a healthy state without alice:

   ```
   2024-06-28 10:22:04,598 ERROR [io.qua.sch.com.run.StatusEmitterInvoker] (vert.x-worker-thread-1) Error occurred while executing task for trigger io.quarkus.quartz.runtime.QuartzSchedulerImpl$QuartzTrigger@3231da4e: java.util.concurrent.CompletionException: java.lang.NullPointerException: Cannot invoke "Object.getClass()" because "entity" is null
        at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:332)
        at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:347)
        at java.base/java.util.concurrent.CompletableFuture.uniWhenComplete(CompletableFuture.java:874)
        at java.base/java.util.concurrent.CompletableFuture.uniWhenCompleteStage(CompletableFuture.java:887)
        at java.base/java.util.concurrent.CompletableFuture.whenComplete(CompletableFuture.java:2357)
        at java.base/java.util.concurrent.CompletableFuture$MinimalStage.whenComplete(CompletableFuture.java:2948)
        at io.quarkus.scheduler.common.runtime.DefaultInvoker.invoke(DefaultInvoker.java:24)
        at io.quarkus.scheduler.common.runtime.StatusEmitterInvoker.invoke(StatusEmitterInvoker.java:35)
        at io.quarkus.quartz.runtime.QuartzSchedulerImpl$InvokerJob$2.call(QuartzSchedulerImpl.java:1101)
        at io.vertx.core.impl.ContextImpl.lambda$executeBlocking$0(ContextImpl.java:177)
        at io.vertx.core.impl.ContextInternal.dispatch(ContextInternal.java:276)
        at io.vertx.core.impl.ContextImpl.lambda$internalExecuteBlocking$2(ContextImpl.java:209)
        at org.jboss.threads.ContextHandler$1.runWith(ContextHandler.java:18)
        at org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2513)
        at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1538)
        at org.jboss.threads.DelegatingRunnable.run(DelegatingRunnable.java:29)
        at org.jboss.threads.ThreadLocalResettingRunnable.run(ThreadLocalResettingRunnable.java:29)
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
        at java.base/java.lang.Thread.run(Thread.java:1570)
   Caused by: java.lang.NullPointerException: Cannot invoke "Object.getClass()" because "entity" is null
        at io.quarkus.hibernate.orm.panache.common.runtime.AbstractJpaOperations.delete(AbstractJpaOperations.java:123)
        at io.quarkus.hibernate.orm.panache.PanacheRepositoryBase.delete(PanacheRepositoryBase.java:83)
        at org.cryptomator.hub.entities.Authority_Repository_ClientProxy.delete(Unknown Source)
        at org.cryptomator.hub.RemoteUserPuller.syncDeletedAuthorities(RemoteUserPuller.java:65)
        at org.cryptomator.hub.RemoteUserPuller.sync(RemoteUserPuller.java:46)
        at org.cryptomator.hub.RemoteUserPuller.sync(RemoteUserPuller.java:40)
        at org.cryptomator.hub.RemoteUserPuller_Subclass.sync$$superforward(Unknown Source)
        at org.cryptomator.hub.RemoteUserPuller_Subclass$$function$$1.apply(Unknown Source)
        at io.quarkus.arc.impl.AroundInvokeInvocationContext.proceed(AroundInvokeInvocationContext.java:73)
        at io.quarkus.arc.impl.AroundInvokeInvocationContext.proceed(AroundInvokeInvocationContext.java:62)
        at io.quarkus.narayana.jta.runtime.interceptor.TransactionalInterceptorBase.invokeInOurTx(TransactionalInterceptorBase.java:136)
        at io.quarkus.narayana.jta.runtime.interceptor.TransactionalInterceptorBase.invokeInOurTx(TransactionalInterceptorBase.java:107)
        at io.quarkus.narayana.jta.runtime.interceptor.TransactionalInterceptorRequired.doIntercept(TransactionalInterceptorRequired.java:38)
        at io.quarkus.narayana.jta.runtime.interceptor.TransactionalInterceptorBase.intercept(TransactionalInterceptorBase.java:61)
        at io.quarkus.narayana.jta.runtime.interceptor.TransactionalInterceptorRequired.intercept(TransactionalInterceptorRequired.java:32)
        at io.quarkus.narayana.jta.runtime.interceptor.TransactionalInterceptorRequired_Bean.intercept(Unknown Source)
        at io.quarkus.arc.impl.InterceptorInvocation.invoke(InterceptorInvocation.java:42)
        at io.quarkus.arc.impl.AroundInvokeInvocationContext.perform(AroundInvokeInvocationContext.java:30)
        at io.quarkus.arc.impl.InvocationContexts.performAroundInvoke(InvocationContexts.java:27)
        at org.cryptomator.hub.RemoteUserPuller_Subclass.sync(Unknown Source)
        at org.cryptomator.hub.RemoteUserPuller.sync(RemoteUserPuller.java:33)
        at org.cryptomator.hub.RemoteUserPuller_ClientProxy.sync(Unknown Source)
        at org.cryptomator.hub.RemoteUserPuller_ScheduledInvoker_sync_5d61926bc8fb63078bd77dc4de24bb346307c2c2.invokeBean(Unknown Source)
        ... 13 more
   ```

1. If I click on the verification icon of my user, the tool-tip is empty:
   
   

2. Why do we not show the Trust Level in the update permission dialog?
  
   

3. As a user without deeper understanding of Hub, I find it a bit confusing why I can not verify all users (account setup not finished)
  
   image

4. Maybe we can add a note where to find the Fingerprint, also it would be nice if the input could be not case sensitive

5. What is the purpose of the tool-tip on hover which only shows the username of the hovered user?
   
   image

6. If `wot_id_verify_len` is set to 0, it is hard to enter the first 0 char :sweat_smile: but if you enter any char and remove it again, the fingerprint is added

image

frontend/src/common/wot.ts
overheadhunter: technically `GET` can be 404 but since we just established a trust chain of length 1 by trusting the user directly, we can expect it to be 200.

Therefore I'd consider an undefined trust as erroneous, so better fail hard.

frontend/src/components/AuditLogDetailsSignedWotId.vue
overheadhunter: Of course we can check the signature internally, I just found it handy for debugging to see the whole raw JWT (but a `console.debug()` would serve this purpose as well)

backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
overheadhunter: I don't think trivial responses like 200 or 404 need to have a description, if it doesn't add any more information. In this case, 200 is the only possible outcome, therefore the description would only be redundant to the one provided in the `@Operation` annotation.

backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
overheadhunter: at least for the 404 case

frontend/src/components/AuditLogDetailsSignedWotId.vue
overheadhunter: changed in 6e10cc0e09402e41fc94503719308329764289f6

SailReal (approved):
LGMT, the only thing I stumbled across was the "Signed Identity" with status "signature invalid"

image

I think users will have problems figuring out what this means, how it can happen, whether it is a problem (red color) and how to solve it. Either we put a tooltip in there or we leave it as it is and explain it in the docs.

tobihagemann (approved):
Functionality is looking good. Approving this since we can make further improvements in a separate PR (or directly on `develop`). I noticed the following:

- Should the "Check Identity" button be removed once the validation was successful (or the trust level is 100%)?
- Is "Unverified" the same as "Trust Level 0%"?
- Why is `wotIdVerifyLen` `@Min(0)`? Shouldn't the minimum be larger? And you've already mentioned it, but I'm not sure if `2` is a good default value. But maybe it's enough? No idea, to be honest.
- Minor UI stuff inside the new popover

overheadhunter:
> I think users will have problems figuring out what this means, how it can happen, whether it is a problem (red color) and how to solve it. Either we put a tooltip in there or we leave it as it is and explain it in the docs.

didn't put much thought into the visual aspects, just made sure the validation runs locally instead of linking to jwt.io



> Should the "Check Identity" button be removed once the validation was successful (or the trust level is 100%)?

I wouldn't remove it, but it should probably change its appearance, depending on whether its a first time verification or revalidation.

> * Is "Unverified" the same as "Trust Level 0%"?

I guess so, where did you find both?

> * Why is `wotIdVerifyLen` `@Min(0)`? Shouldn't the minimum be larger? And you've already mentioned it, but I'm not sure if `2` is a good default value. But maybe it's enough? No idea, to be honest.

I would stick with min 0, as organizations may decide that visual comparison of the ID is sufficient. Default 2 is debatable of course, just found it easier for testing, as I was memorizing the IDs of alice, bob, carol and dave. 😅 

> * Minor UI stuff inside the new popover

Feel free to polish.

PR #308, opened 2025-02-03
Redesign Admin Settings

iammajid:
The layout of the Admin Settings page has been updated to improve usability and scalability as new features are added. Previously, labels were placed above the input fields, but they have now been moved to the left and aligned to the right. This adjustment creates more space for additional descriptions above or below the labels and allows error messages to be displayed more clearly without cluttering the UI.

Since the Admin Settings will continue to grow, this redesign ensures a more structured and readable layout, making future adjustments easier. No functionality has been changed, only the UI layout has been updated.

### DIFF
--- a/frontend/src/components/AdminSettings.vue
+++ b/frontend/src/components/AdminSettings.vue
@@ -20,71 +20,69 @@
         <h3 class="text-lg font-medium leading-6 text-gray-900">
           {{ t('admin.serverInfo.title') }}
         </h3>
-        <p class="mt-1 text-sm text-gray-500 w-full md:w-1/2">
+        <p class="mt-1 text-sm text-gray-500 w-full">
           {{ t('admin.serverInfo.description') }}
         </p>
         <hr class="my-4 pb-6 border-gray-200"/>
-        <div class="md:grid md:grid-cols-3 md:gap-6">
-          <div class="space-y-6 md:col-span-3">
-            <div class="md:grid md:grid-cols-3 md:gap-6">
-              <label for="hubId" class="block text-sm font-medium text-gray-700 md:text-right md:pr-4 md:mt-2">{{ t('admin.serverInfo.hubId.title') }}</label>
-              <div class="mt-1 md:mt-0 md:col-span-2">
-                <div class="md:w-1/2">
-                  <input id="hubId" v-model="admin.hubId" type="text" class="focus:ring-primary focus:border-primary block w-full shadow-xs sm:text-sm border-gray-300 rounded-md bg-gray-200" readonly />
-                </div>
-              </div>
-            </div>
-
-            <div class="md:grid md:grid-cols-3 md:gap-6">
-              <label for="hubVersion" class="block text-sm font-medium text-gray-700 md:text-right md:pr-4 md:mt-2">{{ t('admin.serverInfo.hubVersion.title') }}</label>
-              <div class="mt-1 md:mt-0 md:col-span-2">
-                <div class="md:w-1/2">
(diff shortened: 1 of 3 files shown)


frontend/src/components/AdminSettings.vue
```diff
@@ -17,162 +17,192 @@
 
     <div class="space-y-6 mt-5">
       <div class="bg-white px-4 py-5 shadow-sm sm:rounded-lg sm:p-6">
+        <h3 class="text-lg font-medium leading-6 text-gray-900">
+          {{ t('admin.serverInfo.title') }}
+        </h3>
+        <p class="mt-1 text-sm text-gray-500 w-full md:w-1/2">
```
overheadhunter, line 23: drop the `md:w-1/2`, it looks a bit odd when the text breaks without any space constraints

frontend/src/components/AdminSettings.vue
```diff
@@ -17,162 +17,192 @@
...
-              {{ t('admin.serverInfo.description') }}
-            </p>
-          </div>
-          <div class="mt-5 md:mt-0 md:col-span-2">
-            <div class="grid grid-cols-6 gap-6">
-              <div class="col-span-6 sm:col-span-4">
-                <label for="hubId" class="block text-sm font-medium text-gray-700">{{ t('admin.serverInfo.hubId.title') }}</label>
-                <input id="hubId" v-model="admin.hubId" type="text" class="mt-1 focus:ring-primary focus:border-primary block w-full shadow-xs sm:text-sm border-gray-300 rounded-md bg-gray-200" readonly />
+          <div class="space-y-6 md:col-span-3">
+            <div class="md:grid md:grid-cols-3 md:gap-6">
```
overheadhunter, line 29: grid for a single child.

can be replaced by a simple `<div md:gap-6>`

overheadhunter: even better: use a `<form>` instead of a `<div>`. Even if we don't need to submit anything in this case, it helps to add semantics, as we encapsulate certain inputs that belong together.

overheadhunter (changes requested):
Looks very neat!

Found just one cosmetic problem and a minor HTML structure issue

frontend/src/components/AdminSettings.vue
iammajid: I see your point about md:w-1/2, but I feel that letting the text go full width makes it harder to read. Instead of removing it entirely, I could replace it with max-w-md or max-w-lg, so the text has a natural wrapping point without being too constrained or too wide. What are you thinking?

Screenshot 2025-02-06 at 10 28 55
Screenshot 2025-02-06 at 10 26 56

overheadhunter: If this was a long article, I'd agree to use columns to limit eye movement for easier reading. But this isn't a particularly long text.

If you insist on shortening it, I see two options:

1. Either use the two-column design from the form fields: h3 on the left, p on the right
2. Or shrink the whole container (the white box) on larger screens (maybe add some dummy think on the left, so it resembles the layout of the profile page?)

Both options are merely improvements, though. I'd defer them to a later PR with low priority.

iammajid: Removed in commit 1294c50.

frontend/src/components/AdminSettings.vue
iammajid: Fixed in commit 1294c50.

overheadhunter: same structure should be used for second and third section as well.

iammajid: Fixed in 8970e2c.

overheadhunter (approved):

PR #310, opened 2025-02-04
Feature: User Audit Log

iammajid:
This pull request extends the audit log by adding entries for
* UserAccountReset
* UserKeysChange
* UserSetupCodeChange


frontend/src/i18n/de-DE.json
```diff
@@ -81,7 +81,8 @@
   "auditLog.details.device.register": "Gerät registrieren",
   "auditLog.details.device.remove": "Gerät entfernen",
   "auditLog.details.setting.wot.update": "WoT-Einstellung aktualisieren",
-  "auditLog.details.wot.signedIdentity": "Identität beglaubigt",
+  "auditLog.details.user.account.reset": "Account zurückgesetzt",
+  "auditLog.details.user.account.setup.complete": "Account eingerichtet",
```
SailReal, line 85:
```suggestion
  "auditLog.details.user.account.reset": "Account zurücksetzen",
  "auditLog.details.user.account.setup.complete": "Account einrichten",
```

SailReal (changes requested):
LGTM, only one minor change in the German translation

SailReal (approved):

SailReal:
@iammajid please do not merge until @overheadhunter has approved that the following changes in particular https://github.com/cryptomator/hub/pull/310/commits/132a98b36e93593405b2e3efc1e9336a0fadd42e, https://github.com/cryptomator/hub/pull/310/commits/b5039d0a386014dd93b64e59090528bd9542c0ff and https://github.com/cryptomator/hub/pull/310/commits/fd9357dc4db856a6a7602d27ac8596e60a951af7 are legitimate.

backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
```diff
@@ -79,9 +79,12 @@ public Response putMe(@Nullable @Valid UserDto dto) {
...
 		if (dto != null) {
-			user.setEcdhPublicKey(dto.ecdhPublicKey);
-			user.setEcdsaPublicKey(dto.ecdsaPublicKey);
-			user.setPrivateKeys(dto.privateKeys);
+			if (user.getEcdhPublicKey() == null && user.getEcdsaPublicKey() == null && user.getPrivateKeys() == null) {
+				user.setEcdhPublicKey(dto.ecdhPublicKey);
+				user.setEcdsaPublicKey(dto.ecdsaPublicKey);
+				user.setPrivateKeys(dto.privateKeys);
+				eventLogger.logUserAccountSetupComplete(jwt.getSubject(), jwt.getName());
+			}
```
overheadhunter, line 87: After upgrading from 1.3.x to 1.4.x, each user will get a new ECDSA key on first login.

This is done via `addEcdsaKeyIfMissing()`:

https://github.com/cryptomator/hub/blob/c18af8c3ee69e2ed0c8fa15cf5cd789282ae25ce/frontend/src/common/userdata.ts#L89-L137

I.e. we rely on the ability to update at least this key, even if the other keys are already set. This would be a rather specific condition though, therefore I like it better to allow changing *any* user property when making a PUT request. Makes a clearer API.

overheadhunter: If the sole purpose of the newly added `if` is to log setup completion, I guess it requires something like this:

```java
boolean wasUninitialized = user.getPrivateKeys() == null;
// ...
if (wasUninitialized && user.getPrivateKeys() != null) {
  // log
}
```

frontend/src/components/InitialSetup.vue
```diff
@@ -297,6 +297,7 @@ async function recoverUserKey() {
 }
 
 async function submitBrowserKeys(browserKeys: BrowserKeys, me: UserDto, userKeys: UserKeys) {
+  await backend.users.putMe(me);
```
overheadhunter, line 300: What is the purpose of this change? In some cases, we want a very specific order to avoid data inconsistencies.

overheadhunter (changes requested):

frontend/src/components/InitialSetup.vue
SailReal: In my opinion the order is now more consistent, think about creating a device, then creating the user keys fails (network hiccup), then we have devices in the system without the user keys. Also, in terms of event logging, it looks strange if a device is created first and then the user keys.

backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
SailReal: Agree, with https://github.com/cryptomator/hub/blob/581186fd860df20e2597822f2a47585b38296451/backend/src/main/java/org/cryptomator/hub/api/UsersResource.java#L83-L93 we now log the changes but do not prevent it.

frontend/src/components/InitialSetup.vue
SailReal: As discussed, as we have the account key and we benefit regarding logging here, this is fine.

backend/src/main/java/org/cryptomator/hub/entities/events/EventLogger.java
```diff
@@ -53,6 +53,28 @@ public void logDeviceRemoved(String removedBy, String deviceId) {
...
+
+	public void logUserKeysChanged(String changedBy, String userName) {
+		var event = new UserKeysChangeEvent();
+		event.setTimestamp(Instant.now());
+		event.setChangedBy(changedBy);
+		event.setUserName(userName);
+		auditEventRepository.persist(event);
+	}
+
+	public void logUsersetupCodeChanged(String changedBy) {
```
overheadhunter, line 71:
```suggestion
	public void logUserSetupCodeChanged(String changedBy) {
```

backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
```diff
@@ -79,10 +80,16 @@ public Response putMe(@Nullable @Valid UserDto dto) {
 		user.setPictureUrl(jwt.getClaim("picture"));
 		user.setEmail(jwt.getClaim("email"));
 		if (dto != null) {
-			user.setEcdhPublicKey(dto.ecdhPublicKey);
-			user.setEcdsaPublicKey(dto.ecdsaPublicKey);
-			user.setPrivateKeys(dto.privateKeys);
-			user.setSetupCode(dto.setupCode);
+			if (!Objects.equals(user.getSetupCode(), dto.setupCode)) {
+				user.setSetupCode(dto.setupCode);
+				eventLogger.logUsersetupCodeChanged(jwt.getSubject());
```
overheadhunter, line 85:
```suggestion
				eventLogger.logUserSetupCodeChanged(jwt.getSubject());
```

overheadhunter (approved):
lgtm, but renaming that function wouldn't hurt

PR #313, opened 2025-02-07
Preserve User Language Setting

SailReal:


frontend/src/router/index.ts
```diff
@@ -176,11 +178,15 @@ router.beforeEach((to, from, next) => {
 
 // THIRD check user/browser keys (requires auth)
 router.beforeEach(async (to) => {
+  let me = undefined;
+  if (!to.meta.skipAuth) {
+    me = await userdata.me;
+    i18n.global.locale.value = mapToLocale(me.language);
+  }
```
overheadhunter, line 185: please move this to a separate `beforeEach` section, just for clarity

overheadhunter: (`userdata.me` is memoized, so there should be no performance penalty)

overheadhunter (changes requested):

frontend/src/components/UserProfile.vue
```diff
@@ -99,6 +99,15 @@ async function fetchData() {
   }
 }
 
+async function saveLanguage(locale: Locale) {
+  const updatedUser = me.value;
+  if (updatedUser !== undefined) {
+    updatedUser.language = locale.toString();
```
overheadhunter, line 105: is `locale.toString()` some form of standard? we might be stricter, only allowing iso language codes, so we can rely on the data not having inconsistencies due to strange browsers etc

overheadhunter: nevermind, just saw that Locale is our own enum. We should add an inline comment though to restrict the type (and can then also limit the varchar in the db)

frontend/src/router/index.ts
SailReal: Fixed in [4b749fb](https://github.com/cryptomator/hub/pull/313/commits/4b749fb53cad8f8fd994b01b9d9ed690d4e5ad33)

frontend/src/components/UserProfile.vue
SailReal: Fixed it [e0a09e9](https://github.com/cryptomator/hub/pull/313/commits/e0a09e94da41fdced50a997e7756cf3e80890660)

overheadhunter (approved):

PR #324, opened 2025-02-26
Feature: Add "Browser Language" Option to Language Selection

mindmonk:
Currently, when selecting a language in the user profile, no selection is shown until the user explicitly picks one. In this initial state, the application defaults to the browser's language for localization. Once a language is selected, there is no way to revert to the previous behavior.

With this PR, a new option titled "Browser Language" is added to the language selection list. This ensures that:
- A selection is visible in the dropdown upon first login.
- Users can revert to the browser's language preference after manually selecting another language.

Changes
- Added "Browser Language" as an option in the language selection dropdown.
- Ensured this option is pre-selected if no explicit choice has been made.
- Allowed users to revert back to the browser's language after making a manual selection.


SailReal (changes requested):
> In this initial state, the application defaults to the browser's language for localization. 

This is not true, if the user doesn't have a language selected, English is currently used as default so not set corresponds to English.

SailReal (approved):

PR #332, opened 2025-03-12
Adjusted Column Order in Device List

mindmonk:
The column order in the device list has been rearranged.
Additionally the time format includes now the time of day.

Before:
Before

After:
After

### DIFF
--- a/frontend/src/i18n/index.ts
+++ b/frontend/src/i18n/index.ts
@@ -38,6 +38,11 @@ const defaultShortDatetimeFormat = {
     year: 'numeric',
     month: 'long',
     day: 'numeric'
+  },
+  daytime:{
+    hour: '2-digit', 
+    minute: '2-digit', 
+    second: '2-digit' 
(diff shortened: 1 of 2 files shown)


tobihagemann (approved):

frontend/src/i18n/index.ts
```diff
@@ -38,6 +38,11 @@ const defaultShortDatetimeFormat = {
     year: 'numeric',
     month: 'long',
     day: 'numeric'
+  },
+  daytime:{
+    hour: '2-digit', 
+    minute: '2-digit', 
+    second: '2-digit'
```
SailReal, line 45: Would make a `long` formatter that includes time and date, then you can call `{{ d(device.creationTime, 'long') }}` instead of `{{ d(device.creationTime, 'short') }} {{ d(device.creationTime, 'daytime') }}`

SailReal (approved):

PR #345, opened 2025-06-16
Add BreadcrumbNav component for navigational breadcrumbs

mindmonk:
This PR introduces a new BreadcrumbNav component that provides a flexible and accessible breadcrumb navigation element for the application. It supports both clickable and static breadcrumb items, rendering a separator icon between entries.

### DIFF
--- a/frontend/src/components/BreadcrumbNav.vue
+++ b/frontend/src/components/BreadcrumbNav.vue
@@ -1,0 +1,32 @@
+<template>
+  <nav class="flex items-center text-sm text-gray-500 pb-4" aria-label="Breadcrumb">
+    <ol class="inline-flex items-center">
+      <li v-for="(crumb, index) in crumbs" :key="index" class="flex items-center">
+        <template v-if="crumb.to">
+          <RouterLink :to="crumb.to" class="inline-flex items-center hover:text-gray-700">
+            {{ crumb.label }}
+            <svg v-if="index < crumbs.length - 1" class="w-4 h-4 mx-1 text-gray-400" fill="currentColor" viewBox="0 0 20 20">
+              <path fill-rule="evenodd" d="M7.293 14.707a1 1 0 010-1.414L11.586 9 7.293 4.707a1 1 0 011.414-1.414l5 5a1 1 0 010 1.414l-5 5a1 1 0 01-1.414 0z" clip-rule="evenodd"/>
+            </svg>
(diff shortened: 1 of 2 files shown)


frontend/src/components/BreadcrumbNav.vue
```diff
@@ -0,0 +1,32 @@
+<template>
+  <nav class="flex items-center text-sm text-gray-500 pb-4" aria-label="Breadcrumb">
+    <ol class="inline-flex items-center">
+      <li v-for="(crumb, index) in crumbs" :key="index" class="flex items-center">
+        <template v-if="crumb.to">
+          <RouterLink :to="crumb.to" class="inline-flex items-center hover:text-gray-700">
+            {{ crumb.label }}
+            <svg v-if="index < crumbs.length - 1" class="w-4 h-4 mx-1 text-gray-400" fill="currentColor" viewBox="0 0 20 20">
+              <path fill-rule="evenodd" d="M7.293 14.707a1 1 0 010-1.414L11.586 9 7.293 4.707a1 1 0 011.414-1.414l5 5a1 1 0 010 1.414l-5 5a1 1 0 01-1.414 0z" clip-rule="evenodd"/>
+            </svg>
```
overheadhunter, line 10: maybe we can introduce a separate svg file that contains common paths or shapes that we reuse often?

Then we can make use of [`<use>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Element/use), significantly deduplicating such code in the rendered HTML.

overheadhunter (changes requested):

frontend/src/components/BreadcrumbNav.vue
tobihagemann: I have a different suggestion. Please use an icon from Heroicons, they have something for ChevronRight.

overheadhunter: > I worry about maintainability

Well since SVG is text-based, is can be merged easily via git. With every reusable graphic in a separate line etc.

Furthermore, we can add a Vue component to easily include the corresponding glyph, so virtually no change from Heroicons regarding usability.

overheadhunter: But maybe that's a topic for a separate PR, since it is merely an optimization issue. 😉

overheadhunter (approved):

PR #350, opened 2025-07-31
Add support for regionalized locales and add further langs

SailReal:
When we use the full locale, i18n does the fallback from e.g. de-DE-bavarian --> de-DE --> de, see https://vue-i18n.intlify.dev/guide/essentials/fallback#implicit-fallback-using-locales

This also required to migrate the stored ISO 639‑1 two letter codes to the new full locales.

Also added Latvian, Portuguese (Brazil), Russian, Ukrainian, Chinese (Taiwan)

I also removed `https://cryptomator.org/${languagePathComponent}hub/billing/*` because imo this is getting to complicated and error prone with regionalized locales in comparison to the benefit we get from it.


frontend/src/i18n/index.ts
```diff
@@ -1,36 +1,50 @@
...
+  ITIT = 'it-IT',
+  KOKR = 'ko-KR',
+  LVLV = 'lv-LV',
+  NLNL = 'nl-NL',
+  PTBR = 'pt-BR',
+  PTPT = 'pt-PT',
+  RURU = 'ru-RU',
+  TRTR = 'tr-TR',
+  UKUA = 'uk-UA',
+  ZHTW = 'zh-TW',
```
overheadhunter, line 31: I would prefer `EN_US`, or even violate constant naming rules and use `en_US`.

overheadhunter (commented):
I don't understand the fallback mechanism yet. How does pt-BR know what resources to load when falling back to pt?

frontend/src/i18n/index.ts
SailReal: Tried it before but had linter problems, now not anymore :thinking: , fixed in [c14894b](https://github.com/cryptomator/hub/pull/350/commits/c14894bb06da880d067fad203feec2d8838aaf3c)

overheadhunter (approved):

PR #362, opened 2025-10-17
Security: Pin CI actions with SHA checksum

infeo:
See also https://github.com/cryptomator/cryptomator/pull/4015


SailReal (approved):
LGTM :+1:

PR #371, opened 2025-12-05
Feature: Improve browser locale detection and unify language selection behavior

mindmonk:
This PR improves how the application detects and handles browser languages. The locale normalization logic in i18n/index.ts has been expanded so that short or inconsistent browser locale values (like de from Firefox) are now mapped reliably to full locale codes such as de-DE. A new detectBrowserLocale() helper ensures that the initial i18n locale is always a valid enum value.

In UserProfile.vue, the language selection has been cleaned up by introducing a dedicated languagePreference model and by using the centralized locale detection instead of raw browser values. The browser-language option is now handled explicitly via "browser", and the correct preference is restored when user data is loaded.

Overall, this fixes incorrect language behavior across different browsers and makes the language selection more consistent and predictable.

### DIFF
--- a/frontend/src/components/UserProfile.vue
+++ b/frontend/src/components/UserProfile.vue
@@ -22,7 +22,7 @@
             <ArrowTopRightOnSquareIcon class="-ml-1 mr-2 h-5 w-5" aria-hidden="true" />
             {{ t('userProfile.actions.manageAccount') }}
           </button>
-          <Listbox v-model="locale" as="div">
+          <Listbox v-model="languagePreference" as="div">
             <div class="relative">
               <ListboxButton class="relative w-full inline-flex items-center justify-center px-4 py-2 border border-gray-300 shadow-xs text-sm font-medium rounded-md text-gray-700 bg-white hover:bg-gray-50 focus:outline-hidden focus:ring-2 focus:ring-offset-2 focus:ring-primary">
                 <LanguageIcon class="-ml-1 mr-2 h-5 w-5" aria-hidden="true" />
@@ -33,7 +33,7 @@
               </ListboxButton>
               <transition leave-active-class="transition ease-in duration-100" leave-from-class="opacity-100" leave-to-class="opacity-0">
                 <ListboxOptions class="absolute z-10 mt-1 max-h-60 w-full overflow-auto rounded-md bg-white py-1 shadow-lg ring-1 ring-black/5 focus:outline-hidden text-sm">
-                  <ListboxOption v-slot="{ active, selected }" :value="browserLocale" class="relative cursor-default select-none py-2 pl-3 pr-9 ui-not-active:text-gray-900 ui-active:text-white ui-active:bg-primary" @click="saveLanguage(undefined)">
+                  <ListboxOption v-slot="{ active, selected }" value="browser" class="relative cursor-default select-none py-2 pl-3 pr-9 ui-not-active:text-gray-900 ui-active:text-white ui-active:bg-primary" @click="saveLanguage(undefined)">
                     <span :class="[selected || active ? 'font-semibold' : 'font-normal', 'block truncate']">{{ t('userProfile.actions.changeLanguage.entry.browser') }}</span>
                     <span v-if="selected" :class="[active ? 'text-white' : 'text-primary', 'absolute inset-y-0 right-0 flex items-center pr-4']">
                       <CheckIcon class="h-5 w-5" aria-hidden="true" />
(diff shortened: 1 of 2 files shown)


frontend/src/components/UserProfile.vue
```diff
@@ -33,7 +33,7 @@
               </ListboxButton>
               <transition leave-active-class="transition ease-in duration-100" leave-from-class="opacity-100" leave-to-class="opacity-0">
                 <ListboxOptions class="absolute z-10 mt-1 max-h-60 w-full overflow-auto rounded-md bg-white py-1 shadow-lg ring-1 ring-black/5 focus:outline-hidden text-sm">
-                  <ListboxOption v-slot="{ active, selected }" :value="browserLocale" class="relative cursor-default select-none py-2 pl-3 pr-9 ui-not-active:text-gray-900 ui-active:text-white ui-active:bg-primary" @click="saveLanguage(undefined)">
+                  <ListboxOption v-slot="{ active, selected }" value="browser" class="relative cursor-default select-none py-2 pl-3 pr-9 ui-not-active:text-gray-900 ui-active:text-white ui-active:bg-primary" @click="saveLanguage(undefined)">
```
overheadhunter, line 36: didn't this break model binding?

mindmonk: No, using value="browser" gives the option just a unique value and v-model is changed to languagePreference.

overheadhunter: oh I see this is the option's value. So the db will now contain `locale = 'browser'`. But when loading the site again, where is this value resolved to an actual locale again? I reckon `VueI18n.createI18n({ locale: 'browser', ... })` will fail...?

mindmonk: See line 102-106

frontend/src/i18n/index.ts
```diff
@@ -102,13 +102,55 @@ export const numberFormats: I18nOptions['numberFormats'] = {
...
+
+function getBrowserLocale(): string {
+  if (typeof navigator === 'undefined') {
+    return Locale.EN_US;
+  } else if (navigator.languages && navigator.languages.length > 0) {
+    return navigator.languages[0];
+  } else if (navigator.language) {
+    return navigator.language;
+  }
+  return Locale.EN_US;
```
overheadhunter:
```suggestion
  } else {
    return Locale.EN_US;
  }
```

overheadhunter (changes requested):

frontend/src/components/UserProfile.vue
overheadhunter: This is not what I meant: If the locale is saved in the db and restored in later browser sessions, how will the value 'browser' be turned into the browser's locale (say 'de') before loading the i18n strings?

mindmonk: When “browser” is selected, we don’t store “browser” in the DB, we save 'undefined', just like on first login when the user has not chosen a language yet, and the frontend then falls back to the detected browser locale.

overheadhunter: ah didn't catch the `@click="saveLanguage(undefined)"`.

That isn't rather explicit, though. I can imagine bugs due to the backend treating "undefined" as "don't change anything". Better store a magic constant and resolve it to the browser's locale when loading.

frontend/src/i18n/index.ts
```diff
@@ -102,13 +102,56 @@ export const numberFormats: I18nOptions['numberFormats'] = {
   [Locale.ZH_TW]: defaultNumberFormat
 };
 
-export const mapToLocale = (local: string): Locale =>
-  (Object.values(Locale) as string[]).includes(local)
-    ? (local as Locale)
-    : Locale.EN_US;
+export const mapToLocale = (local: string): Locale => {
```
overheadhunter: can you also fix the typo?


```suggestion
export const mapToLocale = (locale: string): Locale => {
```

overheadhunter (changes requested):
`router/index.ts` calls `mapToLocale(me.language)`, however this function doesn't seem to be prepared to receive `'browser'`

mindmonk:
> `router/index.ts` calls `mapToLocale(me.language)`, however this function doesn't seem to be prepared to receive `'browser'`

@overheadhunter me.language is never set to 'browser'. When the Listbox option “browser” is selected, the value is stored as undefined.

overheadhunter (approved):

Issue #283, opened 2024-06-04
Hide "Archive Vault" button for admin users without ownershop rights


SailReal:
### Please agree to the following

- [X] I have searched [existing issues](https://github.com/cryptomator/hub/issues?q=) for duplicates
- [X] I agree to follow this project's [Code of Conduct](https://github.com/cryptomator/hub/blob/develop/.github/CODE_OF_CONDUCT.md)

### Summary

In the vault details, hide "Archive Vault" button for admin users without ownershop rights

### System Setup

```markdown
- Hub: 1.3.4
- Keycloak: x
- Cryptomator (Desktop/Android/iOS): y
```

### Steps to Reproduce

1. Create user foo
2. Give foo admin privileges
3. Open in Hub vault bar which you can unlock but don't have ownership permission

### Expected Behavior

No "Archive Vault" is displayed.

### Actual Behavior

![image](https://github.com/cryptomator/hub/assets/1786772/105ee570-cce4-4288-b973-b48a6ab8e846)

### Reproducibility

Always

SailReal:
I think we should allow admins to archive all vaults until https://github.com/cryptomator/hub/issues/427 is implemented.

But I think that even if that feature exists, an admin should be able to archive all vaults.

PR #379, opened 2025-12-17
Feature: Hide Archive/Reactivate Vault actions for admins without ownership rights

mindmonk:
**Superseded by https://github.com/cryptomator/hub/pull/430.**

This PR addresses the issue described in #283 by adjusting the visibility logic of the vault action buttons.

The change removes the isAdmin check from the v-if condition controlling the Archive Vault / Reactivate Vault buttons. As a result, these actions are no longer shown to admin users who do not have ownership permissions for the vault.

This is a small UI-only change that:
- Prevents admins without ownership rights from seeing actions they are not allowed to perform
- Avoids failed archive/reactivate attempts that currently result in a 403 Forbidden response from the backend

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -151,6 +151,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Added "Browser Language" option to language selection dropdown, enabling users to revert to browser default language (#324)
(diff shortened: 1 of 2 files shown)


overheadhunter (changes requested):
mention this in CHANGELOG.md

overheadhunter (approved):

PR #404, opened 2026-02-18
Fix null-safe email validation for legacy users

iammajid:
Email validation failed when editing users migrated from legacy systems with empty email. This fix makes email handling null-safe and keeps email required for new users and those with existing email, while making it optional for legacy users without one.

### DIFF
--- a/frontend/src/common/backend.ts
+++ b/frontend/src/common/backend.ts
@@ -87,7 +87,7 @@ export type UserDto = {
   id: string;
   name: string;
   pictureUrl?: string;
-  email?: string | null;
+  email?: string;
   firstName?: string;
   lastName?: string;
   realmRoles: RealmRole[];
--- a/frontend/src/common/backend.ts
+++ b/frontend/src/common/backend.ts
@@ -87,7 +87,7 @@ export type UserDto = {
   id: string;
   name: string;
   pictureUrl?: string;
-  email: string;
+  email?: string | null;
   firstName?: string;
   lastName?: string;
   realmRoles: RealmRole[];
--- a/frontend/src/common/formvalidator.ts
+++ b/frontend/src/common/formvalidator.ts
@@ -103,11 +103,14 @@ export class FormValidator {
   }
 
   /**
-   * Checks if email is required (always for CREATE, only if previously set for EDIT)
+   * Checks if email is required
    */
   static isEmailRequired(isEditMode: boolean, initialEmail?: string): boolean {
-    if (!isEditMode) return true;
-    return !!initialEmail?.trim();
+    if (!isEditMode) {
+      return true; // CREATE: always required
+    }
+    // EDIT: required only if user previously had email
+    return initialEmail !== undefined && initialEmail.trim() !== '';
   }
 
   /**
--- a/frontend/src/common/formvalidator.ts
+++ b/frontend/src/common/formvalidator.ts
@@ -15,11 +15,11 @@ export class FormValidator {
     firstName?: string,
     lastName?: string,
     username: string,
-    email?: string | null,
+    email?: string,
     password: string,
     passwordConfirm: string,
     isEditMode: boolean,
-    initialEmail?: string | null,
+    initialEmail?: string,
     pictureUrl?: string,
     isValidImageUrl?: boolean
   }): ValidationResult {
@@ -97,7 +97,7 @@ export class FormValidator {
   /**
    * Validates email format
    */
-  static isValidEmail(email: string | null | undefined): boolean {
+  static isValidEmail(email?: string): boolean {
     if (!email) return false;
     return /^[^\s@]+@[^\s@]+\.[a-z]{2,}$/i.test(email.trim());
   }
@@ -105,7 +105,7 @@ export class FormValidator {
   /**
    * Checks if email is required (always for CREATE, only if previously set for EDIT)
    */
-  static isEmailRequired(isEditMode: boolean, initialEmail?: string | null): boolean {
+  static isEmailRequired(isEditMode: boolean, initialEmail?: string): boolean {
     if (!isEditMode) return true;
     return !!initialEmail?.trim();
   }
--- a/frontend/src/common/formvalidator.ts
+++ b/frontend/src/common/formvalidator.ts
@@ -15,10 +15,11 @@ export class FormValidator {
     firstName?: string,
     lastName?: string,
     username: string,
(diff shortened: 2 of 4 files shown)


frontend/src/common/backend.ts
```diff
@@ -87,7 +87,7 @@ export type UserDto = {
   id: string;
   name: string;
   pictureUrl?: string;
-  email: string;
+  email?: string | null;
```
overheadhunter: `email?: string` already implies `string | undefined`. Prefer undefined over null

frontend/src/common/formvalidator.ts
```diff
@@ -93,10 +97,19 @@ export class FormValidator {
   /**
    * Validates email format
    */
-  static isValidEmail(email: string): boolean {
+  static isValidEmail(email: string | null | undefined): boolean {
```
overheadhunter: same as above: avoid `null`, no need for `| undefined` when using `?`

```suggestion
  static isValidEmail(email?: string): boolean {
```

overheadhunter (changes requested):
fix optional types (same for `initialEmail`)

frontend/src/common/backend.ts
iammajid: Fixed in 718d612fc55f6c67562a7fcab0549ab2720a4e2e.

frontend/src/common/formvalidator.ts
iammajid: Fixed in 718d612fc55f6c67562a7fcab0549ab2720a4e2e.

frontend/src/components/authority/UserEditCreate.vue
```diff
@@ -393,7 +395,7 @@ async function onSubmit() {
   data.firstName = data.firstName?.trim();
   data.lastName = data.lastName?.trim();
   data.name = data.name.trim();
-  data.email = data.email.trim();
+  data.email = data.email?.trim() ?? '';
```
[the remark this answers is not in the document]
iammajid: Fixed in 9b5b6bb99571f38ed07b0c90192cd5b123f15d9c.

frontend/src/components/authority/UserEditCreate.vue
```diff
@@ -89,9 +89,9 @@
                 {{ t('userEditCreate.email') }}
               </label>
               <div class="mt-1 md:mt-0 md:col-span-2 lg:col-span-1">
-                <input id="email" v-model="data.email" type="email" required :class="[errors.email ? 'border-red-300 focus:border-red-500 focus:ring-red-500' : 'border-gray-300 focus:ring-primary focus:border-primary', 'block w-full max-w-md shadow-sm sm:text-sm rounded-md']"/>
+                <input id="email" v-model="data.email" type="email" :required="emailRequired" :class="[errors.email ? 'border-red-300 focus:border-red-500 focus:ring-red-500' : 'border-gray-300 focus:ring-primary focus:border-primary', 'block w-full max-w-md shadow-sm sm:text-sm rounded-md']"/>
                 <p v-if="errors.email" class="mt-1 text-sm text-red-600">{{ errors.email }}</p>
-                <p v-else-if="data.email && !isValidEmail(data.email?.trim())" class="mt-1 text-sm text-red-600">
+                <p v-else-if="data.email?.trim() && !isValidEmail(data.email)" class="mt-1 text-sm text-red-600">
                   {{ t('userEditCreate.invalidEmail') }}
                 </p>
```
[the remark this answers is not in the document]
iammajid: Fixed in fc9a993b65b382bc30a55e161927c40ced064503.

frontend/src/common/formvalidator.ts
```diff
@@ -93,10 +97,19 @@ export class FormValidator {
...
   }
 
+  /**
+   * Checks if email is required (always for CREATE, only if previously set for EDIT)
+   */
+  static isEmailRequired(isEditMode: boolean, initialEmail?: string): boolean {
+    if (!isEditMode) return true;
+    return !!initialEmail?.trim();
+  }
+
```
overheadhunter, line 117: legibility for humans is more important than compact code (js will be preprocessed during build anyway)

```suggestion
  /**
   * Checks if email is required
   */
  static isEmailRequired(isEditMode: boolean, initialEmail?: string): boolean {
    return isEditMode
      ? initialEmail?.trim() !== '' // required if previously set
      : true; // always required in CREATE mode
  }

```

frontend/src/components/authority/UserEditCreate.vue
```diff
@@ -393,7 +395,7 @@ async function onSubmit() {
   data.firstName = data.firstName?.trim();
   data.lastName = data.lastName?.trim();
   data.name = data.name.trim();
-  data.email = data.email.trim();
+  data.email = data.email?.trim() || undefined;
```
overheadhunter: this is already optional

```suggestion
  data.email = data.email?.trim();
```

overheadhunter (changes requested):

frontend/src/common/formvalidator.ts
iammajid: Fixed in 7e073b7764c7ab8ba3ae9603303f4784f43061a7.

frontend/src/components/authority/UserEditCreate.vue
iammajid: Fixed in 7e073b7764c7ab8ba3ae9603303f4784f43061a7.

frontend/src/common/formvalidator.ts
overheadhunter: My suggestion was also about resolving the negated `if`. If you don't like a ternary operator here, that's fine, but change the order to `if EDIT else CREATE`

overheadhunter: Also, but that's mostly a matter of taste and style: If two branches are mutually exclusive, I prefer an explicit `else` block.

This way you can intuitively grasp the mutually exclusive nature of the branches even without reading the code, just by looking at the structure:

```js
if {
  ██████████
} else {
  █████
}
```

vs

```js
if {
  ██████████
}
█████
```

iammajid: Fixed in b010c55833000d6e8ccfacc36f50d4b82c61b9c9.

Thanks for clarifying, that makes sense! I'll keep that in mind.

overheadhunter (approved):

PR #418, opened 2026-03-11
Fix admin user management: roles, last access, and username field 

iammajid:
Summary                                
  - **Sync realm roles in syncUser()**: \`KeycloakAdminService.syncUser()\` was missing realm role synchronization, causing the admin UI to show no roles   
  for users. Now fetches effective realm-level roles from Keycloak and persists them.
  - **Populate last access data in admin user details**: The admin user detail view showed devices without last access timestamps and IP addresses. Now     
  queries audit events for both modern and legacy devices.                                                                                                  
  - **Disable username field in edit mode**: The username field appeared editable when editing a user, but changes were silently ignored (backend
  \`UpdateUserDto\` has no \`name\` field). The field is now disabled in edit mode.

### DIFF
--- a/backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
+++ b/backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
@@ -375,16 +375,23 @@ public class UsersResource {
 				.map(eva -> VaultResource.VaultDtoWithRole.from(eva.getVault(), eva.getRole()))
 				.toList();
 
-		// Fetch devices (modern devices)
-		Set<DeviceResource.DeviceDto> devices = user.getDevices().stream()
-				.map(DeviceResource.DeviceDto::fromEntity)
-				.collect(Collectors.toSet());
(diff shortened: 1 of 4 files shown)


SailReal (changes requested):
To display devices we have now two database queries with a for each database query in each of the two queries. This is very inefficient.

We need to join the devices with the events in a database query.

overheadhunter:
Please create one PR per topic.

Regarding the last access time: I don't think it is worth the complexity. n+1 queries per user is unacceptable. And even if refactored to a joined query, we would basically need a completely new entity class just this one property. Do we really need this?

SailReal:
@overheadhunter Currently we already show it in the user details page (also pre v1.5.x) so I do not think that we need a new entity class and can just reuse the code

image

iammajid:
Splitting this PR into separate branches. After final alignment this PR will be closed.
- Disable username field in admin user edit mode - #422
- Sync realm roles from Keycloak in admin service - #423
- Last access - #435

PR #420, opened 2026-03-11
Add legacy device deprecation banner for users and admins

mindmonk:
To inform users ahead of time, a warning banner is shown on the **profile page** and **vault list** whenever the logged-in user still has legacy devices. 

Admins additionally see a variant of this banner on the **vault list** and **admin settings page** when any user in the system has legacy devices. 

Screenshots:
User – Profile Page

User – Vault List

Admin – Vault List

Admin – Admin Settings

### DIFF
--- a/backend/src/main/java/org/cryptomator/hub/entities/LegacyDevice.java
+++ b/backend/src/main/java/org/cryptomator/hub/entities/LegacyDevice.java
@@ -98,6 +98,10 @@ public class LegacyDevice {
 			return find("#LegacyDevice.allInList", Parameters.with("ids", ids)).stream();
 		}
 
+		public boolean existsAny() {
+			return count() > 0;
+		}
+
 		public void deleteByOwner(String userId) {
 			delete("#LegacyDevice.deleteByOwner", Parameters.with("userId", userId));
 		}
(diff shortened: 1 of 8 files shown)


overheadhunter (approved):
I have slight concerns regarding fetch efficiency of `userdata.meWithLegacyDevicesAndLastAccess` but since this is just a temporary banner, this should work.

@SailReal do you agree?

backend/src/main/java/org/cryptomator/hub/entities/LegacyDevice.java
```diff
@@ -98,6 +98,10 @@ public Stream<LegacyDevice> findAllInList(List<String> ids) {
 			return find("#LegacyDevice.allInList", Parameters.with("ids", ids)).stream();
 		}
 
+		public boolean existsAny() {
+			return count() > 0;
```
SailReal: `count()` does a `SELECT COUNT(*)` see https://javadoc.io/static/io.quarkus/quarkus-hibernate-orm-panache/3.21.3/io/quarkus/hibernate/orm/panache/PanacheQuery.html#count(). `firstResultOptional` on the other site only queries one item https://javadoc.io/static/io.quarkus/quarkus-hibernate-orm-panache/3.21.3/io/quarkus/hibernate/orm/panache/PanacheQuery.html#firstResultOptional() so we should do


```suggestion
			return findAll().firstResultOptional().isPresent();
```

SailReal (changes requested):
We should remove the banner soon again but then I think it is okay to use `meWithLegacyDevicesAndLastAccess`.

SailReal (approved):

PR #423, opened 2026-03-16
Sync realm roles from Keycloak in admin service

iammajid:
- Sync effective realm roles from Keycloak during user sync
 - Read back effective roles from Keycloak after role update instead of writing submitted roles directly to DB


overheadhunter (changes requested):
We deliberately chose to *first* sync to db and *then* sync to KC.

This ensures consistency: If DB sync fails, no data will reach KC. If KC sync fails, the DB sync will be rolled back (due to the `@Transactional` method finishing exceptionally)

overheadhunter (commented):
What are "effective roles"? Do they include roles inherited from groups?

If yes, should we really attach them to "users" in our db? Doesn't this produce inconsistencies? Or incorrectly sync back to kc users?

infeo:
I added some comments to understand the changes and added unit tests for updating roles

infeo:
Questions asked in https://github.com/cryptomator/hub/pull/423#pullrequestreview-3968035429 actually lead to a redesign.

PR #424, opened 2026-03-18
Feature: responsive pill count in emergency scenario visualisation for mobile

mindmonk:
- `maxVisiblePills` is now a computed value based on container width (2 on mobile < 380px, 3 on desktop)
- Pill width is distributed across fewer pills, making names readable on small screens
- Counter badge (+N) and index filters adapt dynamically to the visible pill count

## Problem
On mobile viewports, the emergency scenario visualization tried to fit 3 pills into a narrow container (~320px), resulting in each pill being only ~68px wide.
User names were almost completely hidden due to truncation.

## Solution
On containers narrower than 380px, only 2 pills are shown (~106px each), with the remaining council members indicated by the existing +N counter badge.

Desktop behavior (3 pills) is unchanged.

### Before
localhost_3000_app_vaults (3)

### After
localhost_3000_app_vaults (4)


overheadhunter:
Please rebase on commit d40ea08, as I eliminated the `calcLeft` stuff

frontend/src/components/emergencyaccess/EmergencyScenarioVisualization.vue
```diff
@@ -64,10 +64,16 @@ let timeoutId: ReturnType<typeof setTimeout> | undefined;
...
-const maxVisibleSlots = 4;
+
+const windowWidth = ref(window.innerWidth);
+const maxUserPills = computed(() => windowWidth.value < 640 ? 2 : 3);
 
 const { selectedUsers, requiredKeyShares } = toRefs(props);
 
+onMounted(() => {
+  window.addEventListener('resize', () => { windowWidth.value = window.innerWidth; });
+});
```
overheadhunter, line 79: instead of observing the window size, observe the outer element's size. this way it will also react to div size changes due to dialog positioning etc.

see https://vueuse.org/core/useResizeObserver/

overheadhunter (changes requested):

frontend/src/components/emergencyaccess/EmergencyScenarioVisualization.vue
mindmonk: We won’t do that as discussed, since it would introduce a new dependency.

frontend/src/components/emergencyaccess/EmergencyScenarioVisualization.vue
```diff
@@ -64,11 +64,22 @@ let timeoutId: ReturnType<typeof setTimeout> | undefined;
...
+const maxUserPills = computed(() => windowWidth.value < 640 ? 2 : 3);
 
 const { selectedUsers, requiredKeyShares } = toRefs(props);
 
+function onWindowResize() {
+  windowWidth.value = window.innerWidth;
+}
+
+onMounted(() => {
+  window.addEventListener('resize', onWindowResize);
```
overheadhunter, line 78: maybe declare [listener to be passive](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#using_passive_listeners)?

```suggestion
  window.addEventListener('resize', onWindowResize { passive: true });
```

mindmonk: passive: true only helps when an event has a default action that the browser would otherwise wait to execute. The resize event is not cancelable (https://developer.mozilla.org/en-US/docs/Web/API/Window/resize_event), meaning preventDefault() has no effect (https://developer.mozilla.org/en-US/docs/Web/API/Event/cancelable), so passive: true has no impact here.

overheadhunter (approved):

PR #425, opened 2026-03-18
Fix false unsaved changes after saving emergency access settings

mindmonk:
## Summary
Fix false "unsaved changes" indicator in admin emergency access settings when "let vault owners choose different key holders" is disabled
## Problem
When saving with `allowChoosingEmergencyCouncil = false`, `defaultMinMembers` is persisted as `selectedUsers.length` instead of `minMembers`. The `hasUnsavedChanges` check compared against `minMembers` (the raw input value), which differed from the saved value — causing the warning to persist immediately after a successful save.                                          

## Solution
Use the same conditional expression in `hasUnsavedChanges` as in the save function: `allowChoosing ? minMembers : selectedUsers.length`


overheadhunter:
> `defaultMinMembers` is persisted as `selectedUsers.length` instead of `minMembers`.

That's correct, because there is no "min members" field visible, when `allowChoosingEmergencyCouncil = false`

frontend/src/components/AdminSettingsEmergencyAccess.vue
```diff
@@ -253,7 +253,7 @@ const hasUnsavedChanges = computed(() => {
   return (
     initialEmergencyAccessSettings.value.enableEmergencyAccess !== enableEmergencyAccess.value ||
     initialEmergencyAccessSettings.value.defaultRequiredEmergencyKeyShares !== requiredShares.value ||
-    initialEmergencyAccessSettings.value.defaultMinMembers !== minMembers.value ||
+    initialEmergencyAccessSettings.value.defaultMinMembers !== (allowChoosing.value ? minMembers.value : selectedUsers.value.length) ||
```
overheadhunter: maybe a simpler way to express this is:

```suggestion
    (allowChoosing.value ? initialEmergencyAccessSettings.value.defaultMinMembers !== minMembers.value : false) ||
```

overheadhunter (approved):

PR #428, opened 2026-03-24
Disabled users should not require a license seat

### LINKED ISSUE #427: Disabled Users should not require a license seat

- [x] I have searched [existing issues](https://github.com/cryptomator/hub/issues?q=) for duplicates
- [x] I agree to follow this project's [Code of Conduct](https://github.com/cryptomator/hub/blob/develop/.github/CODE_OF_CONDUCT.md)

### Summary

Sync disabled User state into Hub, hide those users and do not count them regarding seat count

### Motivation

In Keycloak you can disable users. As a company I want to be able to disable a user so that this user does not require a seat until I active it again (e.g. someone left for a temporary amount of time the company or something like this).

Currently those users can not unlock vaults but otherwise are treated as normal users eventough they can not login to Keycloak so no unlock etc

Image

We just need to
1. add an `enabled` property to our users
2. sync this prop with the syncer using the `enabled` prop of Keycloak's `UserRepresentation` into our database
3. consider this flag during seat calculation

### Considered Alternatives

_No response_

### Anything else?

_No response_

tobihagemann:
Fixes #427

Syncs the `enabled` property from Keycloak's `UserRepresentation` into Hub's user entity. Disabled users are excluded from license seat calculations and hidden from non-admin views (authority search, vault member lists, access grant lists, group member queries). Admins can still see disabled users in the user management page, where they're marked with a "Disabled" badge. Admins can also enable or disable users directly from the user detail page.

Changes:
- New Flyway migration V25 adding `enabled` column to `user_details`
- Keycloak syncer pulls and updates the `enabled` flag on every sync cycle
- All seat-counting queries in `EffectiveVaultAccess` filter on `u.enabled`
- Authority search and vault member endpoints filter out disabled users
- `UserDto` exposes the `enabled` field to the frontend
- Admin user list and user detail pages show a "Disabled" badge
- New `PUT /users/{id}/enabled` admin endpoint to toggle a user's enabled state in Keycloak
- User detail page ellipsis menu shows "Disable" (with confirmation dialog) or "Enable" action

### DIFF
--- a/backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
+++ b/backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
@@ -431,8 +431,8 @@ public class UsersResource {
 	@Transactional
 	@Operation(summary = "enable or disable a user")
 	@APIResponse(responseCode = "204", description = "user updated")
-	public Response setUserEnabled(@PathParam("id") String userId, @Valid @NotNull SetUserEnabledDto dto) {
-		keycloakAdminService.setUserEnabled(userId, dto.enabled());
+	public Response setUserEnabled(@PathParam("id") String userId, boolean enabled) {
+		keycloakAdminService.setUserEnabled(userId, enabled);
 		return Response.noContent().build();
 	}
 
@@ -468,11 +468,6 @@ public class UsersResource {
 			@JsonProperty("password") String password,
 			@JsonProperty("pictureUrl") @Size(max = 255) String pictureUrl,
 			@JsonProperty("realmRoles") @NotNull Set<RealmRole> realmRoles
-	) {
-	}
-
-	public record SetUserEnabledDto(
-			@JsonProperty("enabled") boolean enabled
 	) {
 	}
 
--- a/backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
+++ b/backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
@@ -190,7 +190,7 @@ public class UsersResource {
 		} else {
 			deviceDtos = Set.of();
 		}
-		return new UserDto(user.getId(), user.getName(), user.getPictureUrl(), user.getEmail(), user.getFirstName(), user.getLastName(), user.getLanguage(), Set.of(user.getRealmRoles()), deviceDtos, user.getEcdhPublicKey(), user.getEcdsaPublicKey(), user.getPrivateKeys(), user.getSetupCode());
+		return new UserDto(user.getId(), user.getName(), user.getPictureUrl(), user.getEmail(), user.getFirstName(), user.getLastName(), user.getLanguage(), Set.of(user.getRealmRoles()), user.isEnabled(), deviceDtos, user.getEcdhPublicKey(), user.getEcdsaPublicKey(), user.getPrivateKeys(), user.getSetupCode());
 	}
 
 	/**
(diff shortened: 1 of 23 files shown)


backend/src/main/java/org/cryptomator/hub/keycloak/KeycloakAuthorityProvider.java
```diff
@@ -67,6 +67,7 @@ private KeycloakUserDto mapToUser(UserRepresentation userRepresentation) {
 				userRepresentation.getFirstName(),
 				userRepresentation.getLastName(),
 				pictureUrl,
+				Boolean.TRUE.equals(userRepresentation.isEnabled()),
```
SailReal:
```suggestion
				userRepresentation.isEnabled(),
```

tobihagemann: Fixed in af8aaa50. Also updated the same pattern in KeycloakAdminService.syncUser() for consistency.

backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
```diff
@@ -424,6 +424,18 @@ public UserDto updateUser(@PathParam("id") String userId, @Valid @NotNull Update
 		return UserDto.justPublicInfo(user);
 	}
 
+	@PUT
+	@Path("/{id}/enabled")
+	@RolesAllowed("admin")
+	@Consumes(MediaType.APPLICATION_JSON)
```
SailReal:
```suggestion
	@Consumes(MediaType.TEXT_PLAIN)
```

backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
```diff
@@ -424,6 +424,18 @@ public UserDto updateUser(@PathParam("id") String userId, @Valid @NotNull Update
...
+	@Transactional
+	@Operation(summary = "enable or disable a user")
+	@APIResponse(responseCode = "204", description = "user updated")
+	public Response setUserEnabled(@PathParam("id") String userId, boolean enabled) {
```
SailReal, line 434:
```suggestion
	public Response setUserEnabled(@PathParam("userId") String userId, boolean enabled) {
```

SailReal (changes requested):
Beside the requested changes, it looks good to me 👍

backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
tobihagemann: Fixed in 17f45507.

backend/src/main/java/org/cryptomator/hub/api/UsersResource.java
tobihagemann: The @PathParam value needs to match the {id} path variable in @Path("/{id}/enabled"). Changing it to "userId" would break the endpoint. All sibling endpoints (getUser, updateUser, deleteUser) use the same {id} convention, so keeping it consistent here.

SailReal: Haha yes, to not break the endpoint you could also change `@Path("/{id}/enabled")` to `@Path("/{userId}/enabled")` but I see now where it comes from but I still don't like it, imo `id` is too generic and in other locations we use something like `vaultId` etc but okay, we can refactor it later together with the three other endpoints.

SailReal (approved):

PR #430, opened 2026-03-24
Allow admins to archive and unarchive any vault

### LINKED ISSUE #283: Hide "Archive Vault" button for admin users without ownershop rights

tobihagemann:
Fixes #283.

Admins can view all vaults but previously couldn't archive or unarchive vaults they don't own. The `@VaultRole` annotation's `realmRole` field was shared between the bypass check and the `onMissingVault` check, making it impossible to add admin bypass to the `createOrUpdate` endpoint without breaking vault creation permissions.

This PR introduces a `bypassRealmRole` field on the `@VaultRole` annotation to separate the two concerns. The bypass check is also moved into the vault-exists block so it only applies to existing vaults. Vault creation still requires the `create-vaults` role as before.

Instead of adding admin bypass to the general `createOrUpdate` endpoint (which would let admins modify vault name, description, and emergency key shares), a dedicated `PUT /{vaultId}/archived` endpoint handles archive toggling with its own admin bypass. The `createOrUpdate` endpoint remains restricted to vault owners.

On the frontend, archive and reactivate buttons are now visible to admin users who aren't vault owners. The buttons also appear in the vault recovery section, so admins who are vault owners but have lost their account key can still archive or reactivate. The `setArchived` response normalizes `creationTime` to a `Date` object so the vault detail view renders correctly after toggling.

### DIFF
--- a/backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
+++ b/backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
@@ -513,9 +513,28 @@ public class VaultResource {
 	}
 
 	@PUT
+	@Path("/{vaultId}/archived")
+	@RolesAllowed("user")
+	@VaultRole(value = VaultAccess.Role.OWNER, bypassForRealmRole = true, onMissingVault = VaultRole.OnMissingVault.NOT_FOUND)
+	@Consumes(MediaType.APPLICATION_JSON)
+	@Produces(MediaType.APPLICATION_JSON)
+	@Transactional
+	@Operation(summary = "sets the archived flag of a vault")
+	@APIResponse(responseCode = "200", description = "archived flag updated")
+	@APIResponse(responseCode = "403", description = "requesting user is neither a vault owner nor has the admin role")
+	@APIResponse(responseCode = "404", description = "vault not found")
+	public VaultDto setArchived(@PathParam("vaultId") UUID vaultId, @NotNull Boolean archived) {
+		Vault vault = vaultRepo.findByIdOptional(vaultId).orElseThrow(NotFoundException::new);
+		vault.setArchived(archived);
+		vaultRepo.persistAndFlush(vault);
+		eventLogger.logVaultUpdated(jwt.getSubject(), vault.getId(), vault.getName(), vault.getDescription(), vault.isArchived());
+		return VaultDto.fromEntity(vault);
+	}
+
+	@PUT
 	@Path("/{vaultId}")
 	@RolesAllowed("user") // general authentication. VaultRole filter will check for specific access rights
-	@VaultRole(value = VaultAccess.Role.OWNER, bypassForRealmRole = true, onMissingVault = VaultRole.OnMissingVault.REQUIRE_REALM_ROLE, realmRole = RealmRole.CREATE_VAULTS, bypassForEmergencyAccess = true)
+	@VaultRole(value = VaultAccess.Role.OWNER, onMissingVault = VaultRole.OnMissingVault.REQUIRE_REALM_ROLE, realmRole = RealmRole.CREATE_VAULTS, bypassForEmergencyAccess = true)
 	@Consumes(MediaType.APPLICATION_JSON)
 	@Produces(MediaType.APPLICATION_JSON)
 	@Transactional
--- a/backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
+++ b/backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
@@ -516,7 +516,7 @@ public class VaultResource {
 	@Path("/{vaultId}/archived")
 	@RolesAllowed("user")
 	@VaultRole(value = VaultAccess.Role.OWNER, bypassForRealmRole = true, onMissingVault = VaultRole.OnMissingVault.NOT_FOUND)
-	@Consumes(MediaType.APPLICATION_JSON)
+	@Consumes(MediaType.TEXT_PLAIN)
 	@Produces(MediaType.APPLICATION_JSON)
 	@Transactional
 	@Operation(summary = "sets the archived flag of a vault")
--- a/backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
+++ b/backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
@@ -501,7 +501,7 @@ public class VaultResource {
 	@GET
 	@Path("/{vaultId}")
 	@RolesAllowed("user")
-	@VaultRole(value = {VaultAccess.Role.MEMBER, VaultAccess.Role.OWNER}, bypassForRealmRole = true, realmRole = RealmRole.ADMIN, onMissingVault = VaultRole.OnMissingVault.NOT_FOUND)
+	@VaultRole(value = {VaultAccess.Role.MEMBER, VaultAccess.Role.OWNER}, bypassForRealmRole = true, onMissingVault = VaultRole.OnMissingVault.NOT_FOUND)
 	@Produces(MediaType.APPLICATION_JSON)
 	@Transactional
 	@Operation(summary = "gets a vault")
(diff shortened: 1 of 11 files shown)


backend/src/main/java/org/cryptomator/hub/filters/VaultRoleFilter.java
```diff
@@ -67,22 +62,25 @@ public void filter(ContainerRequestContext requestContext) throws NotFoundExcept
...
+				// user is a member of the emergency access council, so we skip the role check:
+				return;
+			}
 			// check permissions for existing vault:
 			var effectiveRoles = effectiveVaultAccessRepo.listRoles(vaultId, userId);
-			if (Arrays.stream(annotation.value()).noneMatch(effectiveRoles::contains)) {
-				throw new ForbiddenException(forbiddenMsg);
+			var requiredRoles = annotation.value();
+			if (Arrays.stream(requiredRoles).noneMatch(effectiveRoles::contains)) {
+				throw new ForbiddenException("Vault role required: " + Arrays.stream(requiredRoles).map(VaultAccess.Role::name).collect(Collectors.joining(", ")));
```
SailReal, line 78: I know this is not new but just moved but from a security standpoint, do we really want to expose which role is missing?

backend/src/main/java/org/cryptomator/hub/filters/VaultRoleFilter.java
```diff
@@ -67,22 +62,25 @@ public void filter(ContainerRequestContext requestContext) throws NotFoundExcept
...
 			}
 		} else {
 			// how to treat non-existing vault:
 			switch (annotation.onMissingVault()) {
-				case FORBIDDEN -> throw new ForbiddenException(forbiddenMsg);
+				case FORBIDDEN -> throw new ForbiddenException("Vault role required: " + Arrays.stream(annotation.value()).map(VaultAccess.Role::name).collect(Collectors.joining(", ")));
```
SailReal, line 83: Same here like https://github.com/cryptomator/hub/pull/430/changes#r2986376098

SailReal (changes requested):
Hmmmmm in the current implementation an user with the admin role can now edit each vault. I think especially because of https://github.com/cryptomator/hub/blob/0686dcc743a108c9d33cb99a83a49f7fcba6677c/backend/src/main/java/org/cryptomator/hub/api/VaultResource.java#L550 and https://github.com/cryptomator/hub/blob/0686dcc743a108c9d33cb99a83a49f7fcba6677c/backend/src/main/java/org/cryptomator/hub/api/VaultResource.java#L555-L562 this starts to be problematic. Maybe it is time to add next to this huge createOrUpdate thing an "patch" method where we only set specific values etc?

backend/src/main/java/org/cryptomator/hub/filters/VaultRoleFilter.java
tobihagemann: The message string only lives in the Java exception object for server-side logging. `ForbiddenException(String)` constructs a `Response.status(403).build()` with no entity body, so the client just sees a bare 403.

backend/src/main/java/org/cryptomator/hub/filters/VaultRoleFilter.java
tobihagemann: Same as above. The role names never reach the HTTP response body, only server-side debug logs.

frontend/src/common/backend.ts
```diff
@@ -360,6 +360,12 @@ class VaultService {
     return addFallbackPictures ? users.map(fillInMissingPicture) : users;
   }
 
+  public async setArchived(vaultId: string, archived: boolean): Promise<VaultDto> {
+    return axiosAuth.put<VaultDto>(`/vaults/${vaultId}/archived`, archived)
+      .then(response => response.data)
+      .catch((error) => rethrowAndConvertIfExpected(error, 403, 404));
+  }
```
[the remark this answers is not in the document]
tobihagemann: Fixed in 0e760b48.

SailReal (changes requested):
It looks quite good, but if the admin is also owner and has lost the account key, the admin should still be able to archive the vault:

Image

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
```diff
@@ -512,6 +512,25 @@ public VaultDto get(@PathParam("vaultId") UUID vaultId) {
...
+	@Consumes(MediaType.APPLICATION_JSON)
+	@Produces(MediaType.APPLICATION_JSON)
+	@Transactional
+	@Operation(summary = "sets the archived flag of a vault")
+	@APIResponse(responseCode = "200", description = "archived flag updated")
+	@APIResponse(responseCode = "403", description = "requesting user is neither a vault owner nor has the admin role")
+	@APIResponse(responseCode = "404", description = "vault not found")
+	public VaultDto setArchived(@PathParam("vaultId") UUID vaultId, @NotNull Boolean archived) {
+		Vault vault = vaultRepo.findByIdOptional(vaultId).orElseThrow(NotFoundException::new);
+		vault.setArchived(archived);
```
SailReal, line 528: If we unarchive an vault, this can increase the seat count, so we need to check the seat count similar to here

https://github.com/cryptomator/hub/blob/5d5c465caf63e1a3e16c4f91ed93beb7e4747f29/backend/src/main/java/org/cryptomator/hub/api/VaultResource.java#L555-L558

Hmmm but during an update we do not check this as well 🤔https://github.com/cryptomator/hub/blob/5d5c465caf63e1a3e16c4f91ed93beb7e4747f29/backend/src/main/java/org/cryptomator/hub/api/VaultResource.java#L567

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
```diff
@@ -512,6 +512,25 @@ public VaultDto get(@PathParam("vaultId") UUID vaultId) {
 		return VaultDto.fromEntity(vault);
 	}
 
+	@PUT
+	@Path("/{vaultId}/archived")
+	@RolesAllowed("user")
+	@VaultRole(value = VaultAccess.Role.OWNER, bypassForRealmRole = true, onMissingVault = VaultRole.OnMissingVault.NOT_FOUND)
+	@Consumes(MediaType.APPLICATION_JSON)
```
SailReal:
```suggestion
	@Consumes(MediaType.TEXT_PLAIN)
```

SailReal (changes requested):

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
tobihagemann: Fixed in fd9d9392.

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
tobihagemann: Valid point. The same gap exists in the update path of createOrUpdate (line 567), so adding it only here would be inconsistent. Skip for now and follow up later or address now?

SailReal: The implementation would be quite "complex" (we need to check how many seats would be taken when the vault is unarchived and if it still fits into the license) it goes into the direction of this one here

https://github.com/cryptomator/hub/blob/e01e4711403eb8b512953f42c38702f8fe59af3b/backend/src/main/java/org/cryptomator/hub/api/VaultResource.java#L303-L306

and as it is already "broken" in the same way (we did not change any functionality in this regards, still only archiving/unarchiving vaults) IMO we can create an issue with it and address it separately to not increase the complexity of this PR any further.

SailReal: @copilot can you create in issue for this?

SailReal: @coderabbitai if Copilot is not able to create the issue, can you do it?

SailReal: Shame on you @copilot+claude-sonnet-4.6, even @coderabbitai is more powerfull than you on your own platform^^

SailReal (approved):

PR #432, opened 2026-03-27
Check seat count when unarchiving a vault

### LINKED ISSUE #431: Check seat count when unarchiving a vault
## Problem

When unarchiving a vault via `PUT /vaults/{vaultId}/archived` or `vault.setArchived(vaultDto.archived)`, the seat count is not checked against the license limit. Unarchiving a vault restores access for all vault members, which can increase the effective occupied seat count and potentially exceed the licensed seat limit.

This is analogous to the existing check performed in `grantAccess` ([VaultResource.java](https://github.com/cryptomator/hub/blob/e01e4711403eb8b512953f42c38702f8fe59af3b/backend/src/main/java/org/cryptomator/hub/api/VaultResource.java#L303-L306)).

Note: The same gap also exists in `createOrUpdate` when updating an existing vault, but that is pre-existing behavior and not introduced by this PR.

## Acceptance Criteria

- When `PUT /vaults/{vaultId}/archived` is called to **unarchive** a vault or when setting the flag via `vault.setArchived(vaultDto.archived)`, check whether the resulting seat count would exceed the licensed seat limit.
- If the seat limit would be exceeded, reject the request with `402 Payment Required`.
- No seat check is needed when **archiving** a vault (archiving reduces effective seat count).

## Context

Identified during review of PR #430 (Allow admins to archive and unarchive any vault).
See discussion: https://github.com/cryptomator/hub/pull/430#discussion_r2999359502

Raised by: @SailReal

tobihagemann:
When unarchiving a vault, the seat count was not validated against the license limit. Unarchiving restores access for all vault members, which can push the effective seat count beyond the licensed limit.

This adds a seat check on the archive-to-active transition in `setArchived` (`PUT /vaults/{vaultId}/archived`). The check reuses the existing `usersSeatedOnOtherVaults` pattern from `setDirectMembers` to project the seat count after unarchive. If the limit would be exceeded, `PaymentRequiredException` (402) is thrown before any mutation.

`createOrUpdate` (`PUT /vaults/{vaultId}`) no longer handles the `archived` field at all. Archive state is exclusively managed through the dedicated endpoint.

All seat-limit checks across VaultResource now use a shared `ensureSeatsNotExceeded` helper.

Frontend: `ReactivateVaultDialog` has been removed. The reactivation action is now inlined directly in `VaultDetails`, with a 402-specific error message shown inline when the seat limit would be exceeded.

Closes #431.


backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
```diff
@@ -570,6 +566,10 @@ public Response createOrUpdate(@PathParam("vaultId") UUID vaultId, @Valid @NotNu
 		vault.setEmergencyKeyShares(vaultDto.emergencyKeyShares);
 
 		vaultRepo.persistAndFlush(vault); // trigger PersistenceException before we continue with
+		// flush first so countSeatOccupyingUsers sees the unarchived vault; @Transactional rolls back on exception
+		if (wasArchived && !vaultDto.archived) {
+			ensureSeatsNotExceeded(effectiveVaultAccessRepo.countSeatOccupyingUsers());
+		}
```
overheadhunter: We have a different query that simulates seat count instead of looking at the status quo. Need to look it up when I am back on my computer. Just fyi that the "change, check, rollback" approach isn't necessary.

tobihagemann: Fixed in 0cf82325. Added a `countSeatOccupyingUsersIfVaultUnarchived` query that uses `(NOT v.archived OR v.id = :vaultId)` to simulate the seat count before any mutation.

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
```diff
@@ -273,13 +271,10 @@ public Response setDirectMembers(@PathParam("vaultId") UUID vaultId, @NotEmpty M
...
 		var user = userRepo.findByIdOptional(userId).orElseThrow(NotFoundException::new);
-		var usedSeats = effectiveVaultAccessRepo.countSeatOccupyingUsers();
-		if (usedSeats < license.getEntitlements().seats() // free seats available
-				|| effectiveVaultAccessRepo.isUserOccupyingSeat(userId)) { // or user already sitting
-			return addAuthority(vault, user, role);
-		} else {
-			throw new PaymentRequiredException("License seats exceeded. Cannot add more users.");
+		if (!effectiveVaultAccessRepo.isUserOccupyingSeat(userId)) {
+			ensureSeatsNotExceeded(effectiveVaultAccessRepo.countSeatOccupyingUsers() + 1);
 		}
```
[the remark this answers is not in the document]
tobihagemann: Fixed in 24bcf70a. Also applied the same guard to `addGroup`, `setDirectMembers`, and `grantAccess` for consistency.

backend/src/main/java/org/cryptomator/hub/api/VaultResource.java
overheadhunter: This is the check I was talking about:

https://github.com/cryptomator/hub/blob/445fb27041f120d9745ed12e838e7eb5d171cfca/backend/src/main/java/org/cryptomator/hub/api/VaultResource.java#L235-L243

The approach here is "get seat count in *other* vaults + check overlap with *this* vault". The difference of both sets is the number of additional seats required.

Always try reusing existing queries, only add new queries (or logic in general), if efficiency gains are significant. Otherwise, favour code reuse (or even deduplication) for improved maintainability. When in doubt, ask a human 😉

tobihagemann: Damn. I can't read minds after all.

tobihagemann: Fixed in 5708a66d. Replaced `countSeatOccupyingUsersIfVaultUnarchived` with the `usersSeatedOnOtherVaults` pattern and removed the custom query. Also made `createOrUpdate` ignore the archived flag entirely, so unarchiving is only possible through the dedicated endpoint.

frontend/src/components/VaultDetails.vue
```diff
@@ -221,6 +221,15 @@
...
       </div>
+
+      <div v-if="onReactivateError">
+        <p v-if="onReactivateError instanceof PaymentRequiredError" class="text-sm text-red-900 text-right mt-1">
+          {{ t('vaultDetails.error.reactivateFailed') }}
+        </p>
+        <p v-else class="text-sm text-red-900 text-right mt-1">
+          {{ t('common.unexpectedError', [onReactivateError.message]) }}
+        </p>
+      </div>
```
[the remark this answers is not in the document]
tobihagemann: This is done automatically via Crowdin. We only add new localization keys to English so that translators on Crowdin are responsible for the translations. These translations will automatically be pulled into this repository via a Crowdin-synced PR.

frontend/src/components/VaultDetails.vue
```diff
@@ -500,9 +506,17 @@ function showArchiveVaultDialog() {
...
+  try {
+    const vaultDto = await backend.vaults.setArchived(props.vaultId, false);
+    refreshVault(vaultDto);
+  } catch (error) {
+    console.error('Reactivating vault failed.', error);
+    onReactivateError.value = error instanceof Error ? error : new Error('Unknown Error');
+    return;
+  }
+  await refreshLicense();
 }
```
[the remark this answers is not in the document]
tobihagemann: Intentional. If the license refresh fails, the seat count stays stale until the next navigation. Not worth surfacing as an error since the reactivation itself succeeded.

overheadhunter (changes requested):
Introduction of `ensureSeatsNotExceeded` causes unnecessary noise in diff. In order to keep this PR focused on a single change, please undo this (can be done outside this PR, however I don't think it really improves readability)

Rest looks fine.

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Hide Archive/Reactivate Vault actions for admins without ownership rights (#379)
+- Enforce seat limit when unarchiving a vault (#431)
```
overheadhunter: a little more user-centric 😉:

```suggestion
- Check seat limit before unarchiving a vault, ensuring smooth operations without unexpectedly exceeding license limits (#431)
```

overheadhunter (approved):

PR #435, opened 2026-04-01
Populate last access data in admin user details

iammajid:
Summary                                                                                                                                                                                       
                                                                                                                                                                                                
  - Fetch devices with last access data (timestamp + IP) via a single LEFT JOIN query instead of separate queries                                                                               
  - Add findByOwnerWithLastAccess() to Device.Repository and LegacyDevice.Repository                                                                                                            
  - Remove eager device fetching from User.findByIdWithEagerDetails() since devices are now loaded via the joined queries                                                                       
  - No new entity classes reuses existing VaultKeyRetrievedEvent and DeviceDto.fromEntity() overloads


backend/src/main/java/org/cryptomator/hub/entities/Device.java
```diff
@@ -164,6 +164,17 @@ public Stream<Device> findAllInList(List<String> ids) {
 			});
 		}
 
+		public List<Object[]> findByOwnerWithLastAccess(String userId) {
```
overheadhunter: please replace Object[] by a dedicated type (record)

backend/src/main/java/org/cryptomator/hub/entities/User.java
overheadhunter, line 1: please explain this change - seems unrelated and changes behaviour in different places

overheadhunter (changes requested):

backend/src/main/java/org/cryptomator/hub/entities/User.java
iammajid: I initially removed the eager fetch to avoid loading devices twice, once in `findByIdWithEagerDetails` and again via the new `findByOwnerWithLastAccess` queries. But you're right, this method is shared and the change would affect other callers. Reverted.

SailReal:
We should drop this approach with getting the last access time via the audit logs and set a property to the device of the last access.

infeo:
Superseded by https://github.com/cryptomator/hub/pull/466

PR #437, opened 2026-04-01
Truncate long names in detail cards and breadcrumbs

iammajid:
## Summary
  - Add text truncation with hover tooltips to user/group detail cards (UserInfo, GroupInfo) and breadcrumb navigation
  - Extract `displayName` computed property in UserInfo for cleaner template logic
  - Fix email overflow in user detail card with proper flex constraints

### DIFF
--- a/frontend/src/components/authority/UserInfo.vue
+++ b/frontend/src/components/authority/UserInfo.vue
@@ -6,7 +6,7 @@
         <h2 class="text-xl font-semibold text-gray-900 truncate w-full" :title="displayName">
           {{ displayName }}
         </h2>
-        <p v-if="user.firstName || user.lastName" class="text-sm text-gray-500 mt-1 truncate w-full" :title="user.name">
+        <p v-if="displayName !== user.name" class="text-sm text-gray-500 mt-1 truncate w-full" :title="user.name">
           {{ user.name }}
         </p>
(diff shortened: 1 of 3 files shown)


frontend/src/components/authority/UserInfo.vue
```diff
@@ -3,23 +3,18 @@
...
-            {{ user.firstName }} {{ user.lastName }}
-          </template>
-          <template v-else>
-            {{ user.name }}
-          </template>
+        <h2 class="text-xl font-semibold text-gray-900 truncate w-full" :title="displayName">
+          {{ displayName }}
         </h2>
-        <p v-if="user.firstName || user.lastName" class="text-sm text-gray-500 mt-1">
+        <p v-if="user.firstName || user.lastName" class="text-sm text-gray-500 mt-1 truncate w-full" :title="user.name">
```
overheadhunter: the `v-if` seems misplaced here

overheadhunter (changes requested):

frontend/src/components/authority/UserInfo.vue
iammajid: Should the condition be `v-if="displayName !== user.name"` instead, so it relies on the computed property rather than duplicating its logic?

overheadhunter: that should make it more clear, yes

iammajid: Fixed in b3d3828d0b74ad2b129ee70fd0ed0b6f2e447b95.

overheadhunter (approved):

PR #440, opened 2026-04-10
Improve Emergency Access view loading performance  

mindmonk:
### Problem
When opening the Emergency Access view, the frontend made one sequential HTTP request per vault to load its recovery processes (N+1 requests total). This caused unnecessary load time, especially with many vaults.                 

Additionally, the template structure was missing a proper v-if/v-else-if chain, causing other content (license alerts, banners, etc.) to render alongside the loading indicator instead of being hidden until data is ready.    
                                                         
### Solution
A new bulk endpoint loads all recovery processes for the current user's recoverable vaults in a single backend query. On the frontend, this request runs in parallel with the vault list fetch, replacing the sequential loop.      
The template was also restructured so that loading, error, and content states are mutually exclusive.

### DIFF
--- a/backend/src/main/java/org/cryptomator/hub/entities/EmergencyRecoveryProcess.java
+++ b/backend/src/main/java/org/cryptomator/hub/entities/EmergencyRecoveryProcess.java
@@ -36,6 +36,20 @@ import java.util.stream.Stream;
 		"""
 )
 @NamedQuery(
+		name = "EmergencyRecoveryProcess.byCouncilMemberOrProcessMember", query = """
+		SELECT p
+		FROM EmergencyRecoveryProcess p
+		WHERE KEY(p.recoveredKeyShares) = :councilMemberId
+
+		UNION
+
+		SELECT p
+		FROM EmergencyRecoveryProcess p
+		INNER JOIN Vault v ON v.id = p.vaultId
+		WHERE KEY(v.emergencyKeyShares) = :councilMemberId
+		"""
+)
+@NamedQuery(
 		name = "EmergencyRecoveryProcess.byVaultIds", query = """
 		SELECT process
 		FROM EmergencyRecoveryProcess process
@@ -151,6 +165,10 @@ public class EmergencyRecoveryProcess {
 
 		public Stream<EmergencyRecoveryProcess> findByCouncilMember(String councilMemberId) {
 			return find("#EmergencyRecoveryProcess.byCouncilMember", Parameters.with("councilMemberId", councilMemberId)).stream();
+		}
+
+		public Stream<EmergencyRecoveryProcess> findByCouncilMemberOrProcessMember(String councilMemberId) {
+			return find("#EmergencyRecoveryProcess.byCouncilMemberOrProcessMember", Parameters.with("councilMemberId", councilMemberId)).stream();
 		}
 
 		public Stream<EmergencyRecoveryProcess> findByVaultIds(java.util.Collection<UUID> vaultIds) {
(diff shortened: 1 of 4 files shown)


SailReal:
@overheadhunter, do you want to review this PR? You are more familiar with the topic of 'Emergency Access'.

overheadhunter:
XHR looks ok, but the html diff is too much to handle. Can only do manual testing and need to trust the change.

SailReal:
@overheadhunter in such cases, enable "Hide whitespace", then its only a few lines:

image

Resulting in

image

...but make sure to disable it again :sweat_smile:

frontend/src/components/emergencyaccess/EmergencyAccessVaultList.vue
```diff
@@ -310,12 +313,15 @@ async function fetchData() {
...
+        backend.vaults.listRecoverable(),
+        backend.emergencyAccess.findAllProcesses(),
+      ]);
+      vaults.value = fetchedVaults.sort((a, b) => a.name.localeCompare(b.name));
 
+      const processesByVaultId = R.groupBy(allProcesses, p => p.vaultId);
       for (const vault of vaults.value) {
-        const processes = await backend.emergencyAccess.findProcessesForVault(vault.id);
-        vaultRecoveryProcesses.value[vault.id] = processes;
+        vaultRecoveryProcesses.value[vault.id] = processesByVaultId[vault.id] ?? [];
```
overheadhunter: I know it has been this way before, but is it really intended that we load *all* processes, even those where the current user isn't involved?

backend/src/main/java/org/cryptomator/hub/entities/EmergencyRecoveryProcess.java
```diff
@@ -146,6 +167,15 @@ public Stream<EmergencyRecoveryProcess> findByCouncilMember(String councilMember
...
 
+		public Stream<EmergencyRecoveryProcess> findByCouncilMemberOrProcessMember(String councilMemberId) {
+			return find("#EmergencyRecoveryProcess.byCouncilMemberOrProcessMember", Parameters.with("councilMemberId", councilMemberId)).stream();
+		}
+
+		public Stream<EmergencyRecoveryProcess> findByVaultIds(java.util.Collection<UUID> vaultIds) {
+			if (vaultIds.isEmpty()) return Stream.empty();
+			return find("#EmergencyRecoveryProcess.byVaultIds", Parameters.with("vaultIds", vaultIds)).stream();
+		}
+
```
overheadhunter: dead code

backend/src/main/java/org/cryptomator/hub/entities/EmergencyRecoveryProcess.java
```diff
@@ -35,6 +35,27 @@
...
+		WHERE KEY(v.emergencyKeyShares) = :councilMemberId
+		"""
+)
+@NamedQuery(
+		name = "EmergencyRecoveryProcess.byVaultIds", query = """
+		SELECT process
+		FROM EmergencyRecoveryProcess process
+		WHERE process.vaultId IN :vaultIds
+		"""
+)
```
overheadhunter: dead code

frontend/src/components/emergencyaccess/EmergencyAccessVaultList.vue
```diff
@@ -310,12 +312,15 @@ async function fetchData() {
...
       }
```
overheadhunter:
```suggestion
      const processesByVaultId = R.groupBy(allProcesses, p => p.vaultId);
      vaultRecoveryProcesses.value = R.pullObject(vaults.value, v => v.id, v => processesByVaultId[v.id]);
```

overheadhunter (changes requested):

frontend/src/components/emergencyaccess/EmergencyAccessVaultList.vue
mindmonk: see commit: use R.pullObject to build vaultRecoveryProcesses map with fallback

overheadhunter (approved):

PR #454, opened 2026-05-19
Emergency Access UX improvements

mindmonk:
Several small UX improvements across the Emergency Access feature:
- Council members who are not part of an active recovery process can now open the recovery dialog via "Show details" and see the abort button inside. Previously the button was disabled in the emergency access vault list.
- Badges in the vault list are now mutually exclusive: the "insufficient council members" warning is no longer shown in addition to the "broken" badge.
- A "not possible" pill is shown when `requiredKeyShares < 1`.
- The "allow choosing keyholders" checkbox is now also visually rendered as disabled when it is not interactive.


frontend/src/components/emergencyaccess/EmergencyAccessDialog.vue
```diff
@@ -188,7 +188,7 @@
 
               <div class="bg-gray-50 rounded-b-lg px-4 py-3 sm:px-6 sm:flex">
                 <!-- ABORT -->
-                <template v-if="phase !== 'start' && isMeInProcessCouncil && !showSuccess">
+                <template v-if="phase !== 'start' && isMeInCouncil && !showSuccess">
```
overheadhunter: this swaps one condition for the other. now it is no longer possible for *former* council members (who are still part of the process) to see the cancel button

frontend/src/components/emergencyaccess/EmergencyAccessVaultList.vue
```diff
@@ -160,7 +160,7 @@
                       v-if="getProcessByType(vault, 'COUNCIL_CHANGE')"
                       :label="getTypeLabel(vault, 'COUNCIL_CHANGE')"
                       :approval-label="getApprovalLabel(getProcessByType(vault, 'COUNCIL_CHANGE')!)"
-                      :disabled="isBroken(vault) || !isUserInProcessWithType(vault, 'COUNCIL_CHANGE')"
+                      :disabled="isBroken(vault)"
```
overheadhunter, line 163: After this change `isUserInProcessWithType` is now dead code, isn't it?

I have the feeling, that the `!isUserInProcessWithType(vault, '...')` has been there for a purpose, though. Better check twice

overheadhunter (changes requested):
three out of the four changes lgtm.

however, I'm not too sure about the "allow council members to view details" part. please explain the changes.

overheadhunter (approved):

PR #464, opened 2026-06-10
Replace top navbar with sidebar

tobihagemann:
Converts the top navigation bar into a left sidebar.

On desktop it's a full-height sidebar. Administration links (Users, Groups, Audit Logs, Admin) move out of the avatar dropdown into the sidebar under a divider, so navigation lives in one place; the profile menu is pinned to the bottom. On small screens the sidebar becomes an off-canvas drawer behind a hamburger, with the close button at the top-left so it lines up with the hamburger.

The layout is a flexbox app shell (`AppShell.vue`) wrapping `Sidebar` + `SidebarContent`, which `AuthenticatedMain` and `UnlockSuccess` render through.


frontend/src/components/Sidebar.vue
```diff
@@ -0,0 +1,104 @@
...
+          </DialogPanel>
+        </TransitionChild>
+      </div>
+    </Dialog>
+  </TransitionRoot>
+
+  <!-- Static sidebar for desktop -->
+  <div class="group relative hidden bg-tertiary2 transition-[width] duration-200 md:flex md:flex-col" :class="collapsed ? 'md:w-16' : 'md:w-64'">
+    <SidebarContent :me="me" :main-nav="mainNav" :admin-nav="adminNav" :is-admin="isAdmin" :profile-dropdown="profileDropdown" :collapsed="collapsed" />
+    <button type="button" class="absolute top-1/2 -right-3 z-10 flex h-6 w-6 -translate-y-1/2 items-center justify-center rounded-full border border-white/10 bg-tertiary2 text-gray-300 opacity-0 shadow-sm transition-opacity hover:text-white focus:opacity-100 focus:outline-hidden group-hover:opacity-100 group-focus-within:opacity-100" :title="collapsed ? t('nav.sidebar.expand') : t('nav.sidebar.collapse')" @click="toggleCollapsed">
```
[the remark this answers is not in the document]
tobihagemann: Keeping `outline-hidden` here. In Tailwind v4 it renders `outline: 2px solid transparent`, which is the variant that stays visible in forced-colors mode, while `outline-none` sets `outline-style: none` and drops the outline entirely. The toggle also already has `focus:opacity-100` as its focus indicator, so switching to `outline-none` would only remove the forced-colors fallback.

overheadhunter (changes requested):
From UX tests:

1. Does the sidebar really need to be collapsible? Isn't this overengineered? No strong veto, just thinking about future difficulties: Whitelabels would need two logo variants as well (given our own logo redesign, the "logo-text.svg" should go anyway).

2. *if* collapsing is required, the floating chevron button should be replaced by a permanently visible button that's built-into the sidebar. The hover-effect makes me nervous. And the clickable area is way to small, too. I don't want to play sniper.

3. on small screens, when the top bar is visible, the position of "open" (burger menu) and "close" (x button) controls should be identical. I don't want to move the cursor if I reconsidered my "open the menu" decision.

frontend/src/components/SidebarContent.vue
```diff
@@ -0,0 +1,95 @@
...
+          <div v-for="(itemGroup, index) in profileDropdown" :key="`itemGroup-${index}`" class="py-1.5">
+            <router-link v-for="item in itemGroup" :key="item.name" :to="item.to" @click="emit('navigate')">
+              <MenuItem v-slot="{ active }">
+                <div :class="[active ? 'bg-gray-100 text-gray-900' : 'text-gray-700', 'flex items-center px-3.5 py-1.5 text-sm']">
+                  <component :is="item.icon" :class="[active ? 'text-gray-500' : 'text-gray-400', 'flex-none h-5 w-5 mr-3']" aria-hidden="true" />
+                  {{ t(item.name) }}
+                </div>
+              </MenuItem>
+            </router-link>
+          </div>
```
overheadhunter:
```suggestion
          <ul role="list" v-for="(itemGroup, index) in profileDropdown" :key="`itemGroup-${index}`" class="py-1.5">
            <router-link v-for="item in itemGroup" :key="item.name" :to="item.to" @click="emit('navigate')">
              <MenuItem v-slot="{ active }">
                <li :class="[active ? 'bg-gray-100 text-gray-900' : 'text-gray-700', 'flex items-center px-3.5 py-1.5 text-sm']">
                  <component :is="item.icon" :class="[active ? 'text-gray-500' : 'text-gray-400', 'flex-none h-5 w-5 mr-3']" aria-hidden="true" />
                  {{ t(item.name) }}
                </li>
              </MenuItem>
            </router-link>
          </ul>
```

frontend/src/components/Sidebar.vue
overheadhunter, line 1: Instead of rendering everything twice (mobile + desktop), I believe there should be a CSS-based solution without doubling the contents in the DOM.

overheadhunter (changes requested):

frontend/src/components/SidebarContent.vue
tobihagemann: Fixed in b5633c9aa. I left `role="list"` off these `<ul>`s though: they sit inside the Headless UI `MenuItems` popup, and its tree walker only neutralizes role-less elements, so an explicit `role="list"` would survive and leave a stray `list` role inside the `menu`. Bare `<ul>`/`<li>` get `role="none"` and the `MenuItem` stays the `menuitem`.

frontend/src/components/Sidebar.vue
tobihagemann: I looked into collapsing these into one render but kept the two. The mobile drawer leans on the Headless UI Dialog for focus trap, scroll lock, backdrop, and ESC/outside-click close, none of which can be CSS-toggled, so a pure-CSS single render loses them. The only single-render alternative is a matchMedia `v-if`/`v-else`, but that's JS-driven (not the CSS approach you had in mind) and adds remount edge cases for little gain, since this is already just two uses of the shared `SidebarContent` and the mobile copy only mounts while the drawer is open. I'll admit a lot of the Dialog's a11y features matter more on desktop than mobile, but I'd still lean toward keeping Headless UI for it.

overheadhunter (approved):

PR #465, opened 2026-06-15
Remove persistence of Keycloak roles in DB

infeo:
## Decision
This PR removes persisting realm roles in the Hub database. Keycloak (KC) is treated as the single
source of truth and serve roles for display on demand. Roles reported from Keycloak are cached for each user for 30s.

## Context:
* Cryptomator Hub authenticates and authorizes API consumers via Keycloak-issued
JWTs. Realm roles (`user`, `admin`, `create-vaults`) are claims in the access
token, and authorization (`@RolesAllowed`) is enforced entirely from the token
* Hub *also* mirrored each user's realm roles into a
`user_details.realm_roles` column (added in migration `V23`)
* only consumer of the persisted copy was *display*: the `realmRoles` field on
`UserDto`, used only be admins in the user detail page. The data was never used for authorization.
* Keycloak's admin API exposes role mappings per user
  (`/users/{id}/role-mappings/realm`); there is no efficient bulk read

## Pros:
* no realm-role drift between KC and DB, KC is the single source of truth
* less code

## Trade-Offs:
* Admin-only endpoint `GET /users/{id}` to get a single user always includes roles -> one request to KC if  not cached
* if a future feature needs roles for *many* users at once (e.g. a roles column
  in the user list), this decision should be revisited: a per-user read-through
  would become an N+1 against the KC admin API

Supersedes #423


infeo:
@coderabbitai review

backend/src/main/java/org/cryptomator/hub/keycloak/KeycloakAdminService.java
```diff
@@ -315,6 +333,24 @@ public void updateUserRoles(String userId, Set<RealmRole> roles) {
...
+	 * Reads the user's directly assigned realm roles from Keycloak's role-mapping API
+	 * ({@code roles().realmLevel()} — distinct from the {@link UserRepresentation}).
+	 * Roles unknown to {@link RealmRole} are ignored.
+	 *
+	 * @param userId the Keycloak user id
+	 * @return the user's realm role names (see {@link RealmRole#kcName()})
+	 * @throws NotFoundException if no such user exists in Keycloak
+	 */
+	@WithSpan("KeycloakAdminService.realmRolesOf")
+	@CacheResult(cacheName = "realmRoles")
```
overheadhunter: maybe prefix with `keycloak.` to differentiated the origin of cached data (we haven't a naming convention yet, but maybe this is the right time to introduce one)

SailReal (approved):
LGTM 👍!

But I also vote for adding the prefix suggested in https://github.com/cryptomator/hub/pull/465#pullrequestreview-4507139242.

backend/src/main/java/org/cryptomator/hub/keycloak/KeycloakAdminService.java
infeo: Fixed in https://github.com/cryptomator/hub/pull/465/commits/fe8481849e8e60690b83a9e03e9773099a3ce870

PR #468, opened 2026-06-19
Add incremental migration to align with baseline migration

infeo:
Database schema consistency improvements for intended baseline migration:

* Updated the `group_id` and `member_id` columns in the `effective_group_membership` table to use `VARCHAR(255)` with collation "C", matching the `authority.id` column.
* Changed the `device_id` columns in both `audit_event_device_register` and `audit_event_device_remove` tables to `VARCHAR(255)` with collation "C", ensuring their width and collation are consistent with other tables.


overheadhunter:
attempted to save history by changing base branch

PR #485, opened 2026-07-28
Add license leeway

SailReal:
We still have this banner, in case the licence finally expires

but a grace period was added

### DIFF
--- a/backend/src/main/java/org/cryptomator/hub/license/LicenseHolder.java
+++ b/backend/src/main/java/org/cryptomator/hub/license/LicenseHolder.java
@@ -38,7 +38,7 @@ public class LicenseHolder {
 	 * bridging the gap between the end of a billing period and the license server confirming the renewal.
 	 */
 	// visible for testing
-	static final Duration EXPIRATION_LEEWAY = Duration.ofDays(9);
+	static final Duration GRACE_PERIOD = Duration.ofDays(9);
 
 	private final Boolean managedInstance;
 	private final Optional<String> initialId;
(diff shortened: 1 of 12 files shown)


backend/src/main/java/org/cryptomator/hub/license/LicenseHolder.java
```diff
@@ -293,12 +301,30 @@ private LicenseState state() {
...
+
+	/**
+	 * Checks if the license is expired, granting a grace period of {@link #EXPIRATION_LEEWAY}.
 	 *
-	 * @return {@code true}, if the license expired, {@code false} otherwise.
+	 * @return {@code true}, if the license expired more than {@link #EXPIRATION_LEEWAY} ago, {@code false} otherwise.
 	 */
 	public boolean isExpired() {
-		return get().getExpiresAt().toInstant().isBefore(Instant.now());
+		return getLeewayEndsAt().isBefore(Instant.now());
```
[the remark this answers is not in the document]
SailReal: A license without expiration can't reach this code. `get()` only ever returns one of two things: a validated license, and `LicenseValidator.REQUIRED_CLAIMS` includes `exp`, so tokens without it are rejected during `set()`/`loadLicense()`, or the `UnconfiguredLicense` placeholder, which also sets `exp`. So `getExpiresAtAsInstant()` is never `null` here, and the code before this PR relied on the same invariant, `get().getExpiresAt().toInstant()` would have thrown just as much.

The frontend's `expiresAt: Date | null` typing is defensive parsing, not a supported backend state. The server always sends both `expiresAt` and `leewayEndsAt`.

If that invariant ever breaks, I'd rather get a 500 than have `isExpired()` silently return `false` and turn a broken license into one that never expires. There is no such thing as a valid Hub licence that never expires, and there never will be.

frontend/src/common/backend.ts
```diff
@@ -229,18 +229,26 @@ export class LicenseUserInfoDto {
...
 
   public isExpired(): boolean {
-    const now = new Date();
-    return now > (this.expiresAt ?? now); //if expired is null, the license cannot expire
+    return this.leewayEndsAt != null && new Date() > this.leewayEndsAt; // no leeway end means no expiration date, i.e. the license cannot expire
+  }
+
+  public isExpiredWithinLeeway(): boolean {
+    return this.expiresAt != null && new Date() > this.expiresAt && !this.isExpired();
   }
```
overheadhunter: I believe it would be less confusing to have a single method `isExpired(allowLeeway: boolean = false)`.

You would not need knowledge about the different methods to exist.

overheadhunter (changes requested):
Mostly fine. A little api refinement in frontend.

Also, should we call this "leeway" (which suggests to be bidirectional, i.e. `exp ± leeway`)? Or rather something like grace period?

frontend/src/common/backend.ts
SailReal: Right but I do not like calls like `isExpired(true)`. What do you think about an optional string literal `isExpired('allowGracePeriod')`? If you like it as well it is fixed in [33e23e9](https://github.com/cryptomator/hub/pull/485/commits/33e23e92a30a46753b37012a195be6685e238e6d)

SailReal:
Agree, renamed leeway to grace period in [33e23e9](https://github.com/cryptomator/hub/pull/485/commits/33e23e92a30a46753b37012a195be6685e238e6d).

backend/src/main/java/org/cryptomator/hub/license/LicenseHolder.java
```diff
@@ -32,6 +33,13 @@ public class LicenseHolder {
 
 	private static final Logger LOG = Logger.getLogger(LicenseHolder.class);
 
+	/**
+	 * Grace period after the license's expiry date during which the license is still treated as active,
+	 * bridging the gap between the end of a billing period and the license server confirming the renewal.
+	 */
+	// visible for testing
+	static final Duration GRACE_PERIOD = Duration.ofDays(9);
```
overheadhunter, line 41: Let's discuss the length of this period later today.

overheadhunter (approved):
Looks good, but please wait with merging until we talked about the duration.

